PR #4, opened 2020-11-10
LoginForm + RegistrationForm

mattcorner:
# Changes
### LoginForm
* Standard login form for use across apps.
* Added inbuilt validation to form. These are now displayed in the material-ui text field helper text rather than snackbar for a better user experience.
   * All fields are required.
   * Email validation
* Form submit callback is never called while validation fails.
### Select
* Generic select component that would replace Selector.js in NGD, VPM etc. Currently untested but including here because i'd already added it.
### Registration Form
* Standard registration form for use across apps.
* Added inbuilt validation to form. These are now displayed in the material-ui text field helper text rather than snackbar for a better user experience.
   * All fields are required.
   * Email validation
   * Password strength validation using zxcvbn. Min strength required is 3.
   * Password must match
* Form submit callback is never called while validation fails.
### ThemeProvider
* Generic theme provider for material-ui that brings all the custom theming from across our apps into one place.
* Storybook is now wrapped in this ThemeProvider for consistency.


dmbarry86:
When login=true in LoginForm component I think the email and password text fields should be disabled.

dmbarry86:
When loading=true in RegistrationForm component I think all of the text fields should be disabled, just like the button is.

https://deploy-preview-4--ipguk-react-ui.netlify.app/?path=/story/authentication-registrationform--loading

dmbarry86:
Makes it obvious here that the Select used in registration for team has different style to other selects used in the app. The one in registration is outlined, just like all of the other text fields. I think I prefer the outlined fields actually. Thoughts on inconsistency and preferences?

dmbarry86:
Maybe there should be some consistency with the story names. I like the default you have in LoginForm and RegistrationForm. Maybe Copyright and Select could also have default stories instead of primary and normal which they currently have.

dmbarry86:
The doc for onChange callback in Select needs function signature completing, just like how you have completed it for LoginForm showing function syntax and data fields etc. Presumably the selected item is in the callback data.

dmbarry86:
Do we need labelWidth as a property in Select? It seems like the label width is already auto-resizing based on the text, which would be the only behaviour I think you would need?

dmbarry86:
Can the doc be updated so that the props are listed in alphabetical order? Select and RegistrationForm currently in random order. Spotted as I was looking at labelWidth and didn't immediately see label which is up above at the top instead of next to it.

dmbarry86:
Can you update the action which runs tests so that test results are uploaded as artefacts in some nice human readable format? This way I don't have to go delving in to the action steps to find console output for the tests.

dmbarry86 (changes requested):
Some suggestions for improvement have been left as comments.

mattcorner:
> Makes it obvious here that the Select used in registration for team has different style to other selects used in the app. The one in registration is outlined, just like all of the other text fields. I think I prefer the outlined fields actually. Thoughts on inconsistency and preferences?

My personal preference is the standard variant, but i did read something recently that a study had found text fields with an outline were visually better for ux. Google have recently removed standard from their material guidelines in favor of filled it iirc but material-ui have continued to support for legacy. https://material.io/components/text-fields#anatomy

In terms of our adoption, I don't think the mismatch is the worst because they're in totally different app areas, but from a ux point of view, we should maybe consider in future apps.

mattcorner:
> Can you update the action which runs tests so that test results are uploaded as artefacts in some nice human readable format? This way I don't have to go delving in to the action steps to find console output for the tests.

Yeah we can do something like that. What are you hoping to see here past the green status? Would something like this work?

https://github.com/marketplace/actions/jest-annotations-coverage

mattcorner:
> Can the doc be updated so that the props are listed in alphabetical order? Select and RegistrationForm currently in random order. Spotted as I was looking at labelWidth and didn't immediately see label which is up above at the top instead of next to it.

As far as I can tell, the prop types order is taken from the order that they are defined in the Component.propTypes. However, it appears that the input arguments to the first story overwrite this and appear at the top hence the weird order. I'll raise an issue with storybook and see what happens as there are no sort order options available.

dmbarry86:
> > Can you update the action which runs tests so that test results are uploaded as artefacts in some nice human readable format? This way I don't have to go delving in to the action steps to find console output for the tests.
> 
> Yeah we can do something like that. What are you hoping to see here past the green status? Would something like this work?
> 
> https://github.com/marketplace/actions/jest-annotations-coverage

I was thinking of something like a nicely formatted HTML report or similar. Just had a quick look around and this video shows quite nice example of HTML reports for coverage and test results summary. https://www.youtube.com/watch?v=dchwI3Rimxg

https://www.npmjs.com/package/jest-html-reporter

mattcorner:
https://github.com/actions/upload-artifact/issues/14

Issue to track for opening artifacts in browser. Seems to be in the works with no ETA.

dmbarry86:
The new test report is very useful. Probably a bit too useful now because I can see that Select component is missing tests 😉

mattcorner:
https://github.com/storybookjs/storybook/issues/9917 Issue for props table sort function to track

dmbarry86 (approved):

PR #67, opened 2020-12-01
Component/snackbar

### LINKED ISSUE #57: Add Snackbar component
Migrate the Snackbar component currently used in a number of our apps over to this library. Could also consider migrating the snackbar context over as well.

mattcorner:
Snackbar component lifted from NGD. Same as VPM and Model-Manager.

closes #57


dmbarry86:
What are your thoughts about including snackbar context in here as well as I mentioned in https://github.com/IPG-Automotive-UK/react-ui/issues/57? It would provide an example and would also be exactly what we would use in each of our apps where we only want to create one snackbar and then call it from multiple places. At the moment I think we have exactly the same Snackbar context used in VPM & MM.

dmbarry86:
Component itself looks fine.

dmbarry86 (changes requested):
We agreed yesterday to also include the snackbar context provider as it will remove duplication of code across our apps and it is directly related to a component in the library so doesn't feel too out of place.

PR #70, opened 2020-12-02
Component/sidebar

### LINKED ISSUE #56: Add Sidebar & SidebarItem components
Migrate the sidebar components currently used in a number of our apps over to this library. The components in mm can probably be used as a starting point.

mattcorner:
Added 3 new components:
* Sidebar - Sidebar wrapper. Displays logo and title at the top, children in the middle, and app version at the bottom.
* SidebarItem - Sidebar item with Icon, Name, Badge, Expansion from right to left. Nested items are added as children of the component. Children can be nested to an unlimited depth, but nested left padding is only applied to the first level as continued nested padding would take up too much room. 
* SidebarDivider - Divider with specific margins for usage inside a Sidebar.

This should cover all the use cases across the current IPG web tools suite.

Closes #56


dmbarry86:
All looks good apart from the tooltip on the sidebar item. We don't currently have this in any apps anymore and I'm not sure what value it brings having the tooltip say the exact same thing as the name of the item itself. I think we used to have this when we collapsed sidebars down to just icons but don't have it anymore now that the sidebar auto turns in to a menu instead of collapsing. If a tooltip is needed in a particular app then we should just be able to wrap sidebar item with it right? I would suggest removing it from the core component.

Issue #97, opened 2020-12-15
Registration form not empty


dmbarry86:
When loading the registration form it has pre-filled the email and password fields with those I have saved for login. The form should always be empty.

![image](https://user-images.githubusercontent.com/8976377/102205678-6b9e1880-3ec3-11eb-88e5-6bc7644a92d9.png)

dmbarry86:
I suspect the autoComplete attributes are incorrectly set.

PR #99, opened 2020-12-15
Bug/registration form auto complete

dmbarry86:
## Summary
Fixed issue with registration form auto-completing email and password for the new user with those of a current user. Also noticed that the autocomplete attribute was incorrectly set on some of the password fields user in the password change components so I have updated those at the same time.

Closes https://github.com/IPG-Automotive-UK/react-ui/issues/97

### Before
image

### After
image


mattcorner (approved):
LGTM. Updated release notes (bumped to 1.6.0 as it contains a new feature that is unreleased at the moment).

Will leave release to you.

PR #100, opened 2020-12-16
Bug/invalid select value

dmbarry86:
Catch the case where invalid value is passed in and just display empty in this case as the underlying select component requires. An invalid value is one that is not a member of the options.

Closes https://github.com/IPG-Automotive-UK/ngd/issues/238

### DIFF
--- a/src/LoginForm/LoginForm.test.js
+++ b/src/LoginForm/LoginForm.test.js
@@ -1,4 +1,10 @@
-import { act, fireEvent, render, screen } from "@testing-library/react";
+import {
+  act,
+  fireEvent,
+  render,
+  screen,
+  waitFor
+} from "@testing-library/react";
 import LoginForm from "./";
 import React from "react";
 import userEvent from "@testing-library/user-event";
@@ -24,15 +30,15 @@ describe("LoginForm", () => {
   it("returns form information to callback when successfully validated", async () => {
     const onLogin = jest.fn(data => data);
     const elements = setup({ onLogin });
-    await act(async () => {
-      userEvent.type(elements.inputs.email, "joe.bloggs@domain.com");
-      userEvent.type(elements.inputs.password, "indigo shark wallplug");
-      fireEvent.submit(elements.submit);
-    });
-    expect(onLogin).toHaveReturnedWith({
-      email: "joe.bloggs@domain.com",
-      password: "indigo shark wallplug"
-    });
+    userEvent.type(elements.inputs.email, "joe.bloggs@domain.com");
+    userEvent.type(elements.inputs.password, "indigo shark wallplug");
+    fireEvent.submit(elements.submit);
+    await waitFor(() =>
+      expect(onLogin).toHaveReturnedWith({
+        email: "joe.bloggs@domain.com",
+        password: "indigo shark wallplug"
+      })
+    );
   });
   it("doesnt call callback on validation errors", async () => {
     const onLogin = jest.fn();
--- a/src/LoginForm/LoginForm.test.js
+++ b/src/LoginForm/LoginForm.test.js
@@ -41,4 +41,14 @@ describe("LoginForm", () => {
     fireEvent.submit(elements.submit);
     await waitFor(() => expect(onLogin).not.toHaveBeenCalled());
   });
+  it("displays error message to user on validation fail", async () => {
+    const elements = setup();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
+    fireEvent.submit(elements.submit);
+    await waitFor(() =>
+      expect(
+        screen.queryByText("Please enter a valid email address")
+      ).toBeInTheDocument()
+    );
+  });
 });
--- a/src/LoginForm/LoginForm.test.js
+++ b/src/LoginForm/LoginForm.test.js
@@ -1,10 +1,4 @@
-import {
-  act,
-  fireEvent,
-  render,
-  screen,
-  waitFor
-} from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import LoginForm from "./";
 import React from "react";
 import userEvent from "@testing-library/user-event";
@@ -43,11 +37,8 @@ describe("LoginForm", () => {
   it("doesnt call callback on validation errors", async () => {
     const onLogin = jest.fn();
     const elements = setup({ onLogin });
-    await act(async () => {
-      userEvent.type(elements.inputs.email, "joe.bloggs");
-      // missing password
-      fireEvent.submit(elements.submit);
-    });
-    expect(onLogin).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onLogin).not.toHaveBeenCalled());
   });
 });
--- a/src/PasswordChangeDialog/PasswordChangeDialog.test.js
+++ b/src/PasswordChangeDialog/PasswordChangeDialog.test.js
@@ -1,4 +1,4 @@
-import { act, fireEvent, render, screen } from "@testing-library/react";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import PasswordChangeDialog from "./";
 import React from "react";
 import userEvent from "@testing-library/user-event";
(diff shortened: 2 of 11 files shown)


src/Select/Select.test.js
```diff
@@ -49,4 +49,18 @@ describe("Select", () => {
     const inputBase = container.querySelector(".MuiInputBase-root");
     expect(inputBase).toHaveClass("Mui-error");
   });
+  test("display empty if value is not a valid option", async () => {
```
mattcorner, line 52: These tests don't need to be async as they don't await anything.

mattcorner:
```suggestion
  test("display empty if value is not a valid option", () => {
```

src/Select/Select.test.js
```diff
@@ -49,4 +49,18 @@ describe("Select", () => {
...
+    const { container } = render(
+      <SelectWithState options={options} value="Invalid value" />
+    );
+    const baseComponent = container.querySelector(".MuiSelect-nativeInput");
+    expect(baseComponent.value).toBe("");
+  });
+  test("display the value if it's a valid option", async () => {
```
mattcorner, line 59:
```suggestion
  test("display the value if it's a valid option", () => {
```

mattcorner (changes requested):
Remove async from tests and its good to go.

dmbarry86:
I've removed the async from the new tests. I've also added an eslint rule to catch such cases in the future. In adding the rule I found some existing code which was violating the rule so I have fixed those as well. This then uncovered some tests which were never going to fail so I fixed those as well.

src/LoginForm/LoginForm.test.js
```diff
@@ -24,24 +24,21 @@ describe("LoginForm", () => {
...
     const elements = setup({ onLogin });
-    await act(async () => {
-      userEvent.type(elements.inputs.email, "joe.bloggs");
-      // missing password
-      fireEvent.submit(elements.submit);
-    });
-    expect(onLogin).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onLogin).not.toHaveBeenCalled());
```
mattcorner, line 42: This should now be testing for the error message on screen as we discovered. onLogin is not called by default.

src/PasswordChangeDialog/PasswordChangeDialog.test.js
```diff
@@ -33,28 +33,26 @@ describe("PasswordChangeDialog", () => {
...
     const onSubmit = jest.fn();
     const elements = setup({ onSubmit });
-    await act(async () => {
-      userEvent.type(elements.inputs.newPassword, "abc123"); // top 100 password
-      fireEvent.submit(elements.submit);
-    });
-    expect(onSubmit).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.newPassword, "abc123"); // top 100 password
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onSubmit).not.toHaveBeenCalled());
```
mattcorner, line 56: Test should be testing for error message on screen rather than lack of callback execution.

src/PasswordChangeForm/PasswordChangeForm.test.js
```diff
@@ -26,43 +26,43 @@ describe("PasswordChangeForm", () => {
...
     const onSubmit = jest.fn();
     const elements = setup({ onSubmit });
-    await act(async () => {
-      userEvent.type(elements.inputs.password, "abc123"); // common password
-      fireEvent.submit(elements.submit);
-    });
-    expect(onSubmit).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.password, "abc123"); // common password
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onSubmit).not.toHaveBeenCalled());
```
mattcorner, line 44: Change to test for error on screen.

src/PasswordChangeForm/PasswordChangeForm.test.js
```diff
@@ -26,43 +26,43 @@ describe("PasswordChangeForm", () => {
...
-      await act(async () => {
-        userEvent.type(elements.inputs.password, "something");
-      });
-      expect(
-        screen.findByText("Password strength: 0/4. Minimum required 3+.")
-      ).toBeTruthy();
+      userEvent.type(elements.inputs.password, "something");
+      await waitFor(() =>
+        expect(
+          screen.findByText("Password strength: 0/4. Minimum required 3+.")
```
mattcorner, line 52: should be queryByText and toBeInTheDocument()

src/PasswordChangeForm/PasswordChangeForm.test.js
```diff
@@ -26,43 +26,43 @@ describe("PasswordChangeForm", () => {
...
-      await act(async () => {
-        userEvent.type(elements.inputs.password, "something");
-      });
-      expect(
-        screen.findByText("Add another word or two. Uncommon words are better.")
-      ).toBeTruthy();
+      userEvent.type(elements.inputs.password, "something");
+      await waitFor(() =>
+        expect(
+          screen.findByText(
```
mattcorner, line 61: should be queryByText and toBeInTheDocument()

src/RegistrationForm/RegistrationForm.test.js
```diff
@@ -34,36 +34,32 @@ describe("RegistrationForm", () => {
...
-      // incorrect email format
-      // missing first, lastname, team, password + password repeat
-      fireEvent.submit(elements.submit);
-    });
-    expect(onRegister).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.email, "joe.bloggs");
+    // incorrect email format
+    // missing first, lastname, team, password + password repeat
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onRegister).not.toHaveBeenCalled());
```
mattcorner, line 62: Test for error message on screen instead

mattcorner (changes requested):
Some changes needed now that we've learned a bit more about the tests. I'll action as they're from me.

mattcorner (approved):
LGTM (finally)

src/LoginForm/LoginForm.test.js
```diff
@@ -41,4 +41,14 @@ describe("LoginForm", () => {
     fireEvent.submit(elements.submit);
     await waitFor(() => expect(onLogin).not.toHaveBeenCalled());
   });
+  it("displays error message to user on validation fail", async () => {
+    const elements = setup();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
```
dmbarry86, line 46: shouldn't userEvent be wrapped in act @mattcorner ?

src/PasswordResetForm/PasswordResetForm.test.js
```diff
@@ -42,4 +42,14 @@ describe("PasswordResetForm", () => {
       ).toBeInTheDocument()
     );
   });
+  it("displays error message to user on validation fail", async () => {
+    const elements = setup();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
```
dmbarry86, line 47: shouldn't userEvent be wrapped in act @mattcorner ?

mattcorner: You'd think so, but not in this case because the form validation is only applied onSubmit, not onTouch like it is in something like RegistrationForm.

src/LoginForm/LoginForm.test.js
mattcorner: You'd think so, but not in this case because the form validation is only applied onSubmit, not onTouch like it is in something like RegistrationForm.

src/LoginForm/LoginForm.test.js
```diff
@@ -24,24 +24,31 @@ describe("LoginForm", () => {
...
-      fireEvent.submit(elements.submit);
-    });
-    expect(onLogin).not.toHaveBeenCalled();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
+    fireEvent.submit(elements.submit);
+    await waitFor(() => expect(onLogin).not.toHaveBeenCalled());
+  });
+  it("displays error message to user on validation fail", async () => {
+    const elements = setup();
+    userEvent.type(elements.inputs.email, "joe.bloggs"); // missing password
```
dmbarry86, line 46: It's wrong email not missing password as the comment suggests

src/PasswordResetForm/PasswordResetForm.test.js
dmbarry86: Also, it's testing wrong email not missing password as the comment suggests. Don't think there is a password on this form.

src/LoginForm/LoginForm.test.js
dmbarry86: Also, it's testing wrong email not missing password as the comment suggests. Password is included on this form but that's not what you are testing for.

dmbarry86: > You'd think so, but not in this case because the form validation is only applied onSubmit, not onTouch like it is in something like RegistrationForm.

So we're saying the rule is only wrap userEvent with act if the userEvent call triggers some async code from the react-hook-form library which updates the DOM? In all other cases you don't need to wrap userEvent?

Issue #98, opened 2020-12-15
Change password components don't set username in password manager


dmbarry86:
If using a password manager, e.g. Chrome, the username field is not set when using one of the password change components. This can cause user to change password and think they have updated their password manager but actually it will have created a new entry with empty username.

![image](https://user-images.githubusercontent.com/8976377/102231599-0d366180-3ee6-11eb-8cd2-3f74e4decd43.png)

![image](https://user-images.githubusercontent.com/8976377/102231567-03146300-3ee6-11eb-85e0-da2556112b7f.png)

dmbarry86:
At the same time, check that the registration form is correctly setting the username field for use in password managers. Just had a quick bash and it seemed to pick up the surname field as the username with Chrome password manager.

PR #101, opened 2020-12-17
Password manager integration

dmbarry86:
Fixes issue with password manager not always recognising username field when registering a new account, changing password when logged in or resetting password. Fix was actually a simple one in the end and just required autocomplete on the email fields to be set to username. Have tested all password components using Chrome & Edge password managers and now username (email address in our case) is correctly picked up.

Closes https://github.com/IPG-Automotive-UK/react-ui/issues/98

### DIFF
--- a/src/LoginForm/LoginForm.js
+++ b/src/LoginForm/LoginForm.js
@@ -27,7 +27,7 @@ export default function LoginForm({ loading = false, onLogin = () => {} }) {
             fullWidth
             label="Email Address"
             name="email"
-            autoComplete="email"
+            autoComplete="username"
             autoFocus={!loading}
             inputProps={{ "aria-label": "email" }}
             inputRef={register({
--- a/src/RegistrationForm/RegistrationForm.js
+++ b/src/RegistrationForm/RegistrationForm.js
@@ -83,7 +83,7 @@ export default function RegistrationForm({
             fullWidth
             label="Email Address"
             name="email"
-            autoComplete="email"
+            autoComplete="username"
             inputProps={{ "aria-label": "email" }}
             inputRef={register({
               pattern: {


mattcorner:
Changes look good for the registration and login forms, tested in chrome. 

 I don't think it closes the linked issue which talks about the password change forms though. We discussed that they probably need a hidden username/email field

dmbarry86:
> Changes look good for the registration and login forms, tested in chrome.
> 
> I don't think it closes the linked issue which talks about the password change forms though. We discussed that they probably need a hidden username/email field to allow the password manager to know which credentials are being modified and store them.

It certainly has fixed it for me. The password manager now correctly knows which user you are logged in as and when I change password it now correctly offers to update the already saved password for the currently logged in user. I assumed this was because the login form now correctly knows about username and password manager then makes the link when changing password. I tested changing password in both Chrome and Edge and it works fine. Is it not working for you?

dmbarry86:
> Changes look good for the registration and login forms, tested in chrome.
> 
> I don't think it closes the linked issue which talks about the password change forms though. We discussed that they probably need a hidden username/email field to allow the password manager to know which credentials are being modified and store them. I guess maybe it works when there's only one account for a domain, but what if you have two accounts?

No it works fine for me with multiple accounts saved.

dmbarry86:
> I guess maybe it works when there's only one account for a domain, but what if you have two accounts?

Works fine for me with multiple accounts

mattcorner (approved):
Yeah okay, the inbuilt password managers in Chrome and Edge seem to be a cleverer than LastPass in this regard. It seems they default to updating the last credentials used for a domain which makes sense, whereas LastPass clocks the credentials change, but asks me which of my accounts I want to update on this domain.

PR #119, opened 2021-02-09
Added SearchBar Component

### LINKED ISSUE #54: Add search component
Migrate the search component currently used in a number of our apps to this library.

image

jegasriskantha:
# Changes: 

- Added SearchBar component to react-ui

# Notes
- This will need to me merged before merging the associated NGD pull request. 

closes #54

### DIFF
--- a/src/SearchBar/SearchBar.js
+++ b/src/SearchBar/SearchBar.js
@@ -1,0 +1,81 @@
+import { IconButton, InputBase, Paper, makeStyles } from "@material-ui/core";
+import CloseIcon from "@material-ui/icons/Close";
+import PropTypes from "prop-types";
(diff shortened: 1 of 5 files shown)


dmbarry86:
The value control in Storybook doesn't seem to be working. Would expect to be able to define the value prop and then it show up as the value (like I can do for the custom placeholder prop which does work). We want this component to be a controlled component, i.e. we can control the value and synch up with some parent state which is linked to the callback.

value control

dmbarry86 (changes requested):
The component and tests look great 👍

Just the question about the value prop in the story to check.

dmbarry86 (approved):
All good

PR #129, opened 2021-03-03
Update storybook stories to update dynamic values

jegasriskantha:
Updated the following  Storybook stories to ensure that the controls work as expected in storybook: 

- SwitchField
- Snackbar
- Select
- PasswordChangeDialog

### DIFF
--- a/src/PasswordChangeDialog/PasswordChangeDialog.stories.js
+++ b/src/PasswordChangeDialog/PasswordChangeDialog.stories.js
@@ -10,6 +10,11 @@ export default {
 
 const Template = args => {
   const [open, setOpen] = React.useState(false);
+
+  React.useEffect(() => {
+    setOpen(args.open);
+  }, [args.open]);
+
   return (
     <>
       <Button onClick={() => setOpen(true)}>Open dialog</Button>
--- a/src/Select/Select.stories.js
+++ b/src/Select/Select.stories.js
@@ -3,6 +3,9 @@ import React from "react";
 import Select from "./Select";
 
 export default {
+  argTypes: {
+    value: { type: "string" }
+  },
   component: Select,
   title: "General/Select"
 };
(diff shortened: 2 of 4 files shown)


dmbarry86:
Changing the value control for Select component doesn't seem to have any effect. Would expect the value to be the selected item.

dmbarry86:
PasswordChangeDialog looks good

dmbarry86:
The open control for Snackbar doesn't seem to have any effect in any of the 3 stories. Other controls for Snackbar seem fine.

dmbarry86:
SwitchField looks good

dmbarry86 (changes requested):
Select and Snackbar components still not quite right. See comments thread.

jegasriskantha:
> The open control for Snackbar doesn't seem to have any effect in any of the 3 stories. Other controls for Snackbar seem fine.

Fixed this now. I also changed the default for each of the stories so that the snack bar is closed initially.

jegasriskantha:
> Changing the value control for Select component doesn't seem to have any effect. Would expect the value to be the selected item.

I think this works as expected. Would you expect not to have to enter the value within quotations in Storybook? 

Select value control

dmbarry86:
> > Changing the value control for Select component doesn't seem to have any effect. Would expect the value to be the selected item.
> 
> I think this works as expected. Would you expect not to have to enter the value within quotations in Storybook?
> 
> Select value control

I don't think I would expect quotes. The options prop doesn't include quotes so I would kind of expect to pass in a corresponding value to select one of them.

dmbarry86:
Same with all of the other string realted props actually, none of them include quotes.

image

dmbarry86 (changes requested):
Snackbar is good now. Select still not consistent.

jegasriskantha:
> Snackbar is good now. Select still not consistent.

Fixed. Updated the propType for Storybook so it now knows to treat "value" as a string. 
Select value control fixed

dmbarry86 (approved):

PR #132, opened 2021-03-08
Add error prop to Select component

dmbarry86:
Added new error prop to the Select componet so it can be controlled outside the component. Prior to this the component itself had some built-in logic to decide when to show the error state. This obviously isn't that useful if you want to control the error state yourself in an app (as I do).


jegasriskantha (approved):
Looks good to me.

PR #229, opened 2021-09-14
#405 add no license page

evoinea:


src/NoLicense/NoLicense.js
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Message to display below the header text
+   */
+  labelBody: PropTypes.string,
+
+  /**
+   * Message to diplay alongside the icon (header)
+   */
+  labelHeader: PropTypes.string
+};
```
dmbarry86, line 38: I think it would be good if we could try to stick to single word prop names. Maybe instead of labelHeader and labelBody we could use title and description.

dmbarry86 (changes requested):
Component looks good 👍

Just prop names to change to single word to make it easier to use.

src/NoLicense/NoLicense.js
evoinea: done

dmbarry86 (approved):

PR #257, opened 2021-11-08
Feature/autocomplete component

dmbarry86:
Added a new Autocomplete component for use with ID application. The component syntax is identical to the existing Select component but allows you to type in the text field to autocomplete/filter the available options.

image


jegasriskantha (approved):
Looks good to me

PR #264, opened 2021-11-15
Feature/slider component

evoinea:
Contributes to  https://github.com/IPG-Automotive-UK/instrument-designer/issues/4.

Added a new Slider component for use with Instrument designer app.
image

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -10,7 +10,7 @@ export default function Slider({
   max = 10,
   min = 1,
   onChange = () => {},
-  onChangeComitted = () => {},
+  onChangeCommitted = () => {},
   orientation = "horizontal",
   predefValues = [],
   labels = true,
@@ -60,7 +60,7 @@ export default function Slider({
           min={min}
           max={max}
           onChange={onChange}
-          onChangeComitted={onChangeComitted}
+          onChangeCommitted={onChangeCommitted}
           value={typeof value !== "undefined" ? value : null}
           orientation={orientation}
         />
@@ -72,7 +72,7 @@ export default function Slider({
           min={min}
           max={max}
           onChange={onChange}
-          onChangeComitted={onChangeComitted}
+          onChangeCommitted={onChangeCommitted}
           value={typeof value !== "undefined" ? value : null}
           orientation={orientation}
         />
@@ -124,7 +124,7 @@ Slider.propTypes = {
    * _event_: The event source of the callback.
    * _value_: The new value.
    */
-  onChangeComitted: PropTypes.func,
+  onChangeCommitted: PropTypes.func,
   /**
    * Slider orientation
    */
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -19,13 +19,13 @@ export default function Slider({
   title,
   value
 }) {
-  // add ticks labels if they are not pre defined
+  // get ticks labels if they are not pre defined
   const range = (start, stop, step) =>
     Array.from(
       { length: (stop - start) / step + 1 },
       (_, i) => start + i * step
     );
-
+  // show/hide labels
   let marks = [];
   let stepSize = step;
   if (labels) {
@@ -41,6 +41,7 @@ export default function Slider({
   } else {
     marks = false;
   }
+  // return components
   return (
     <Box>
       <Typography>{title}</Typography>
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -6,14 +6,14 @@ import React from "react";
  * Discrete slider component
  */
 export default function Slider({
+  labelPosition = "bottom",
+  labels = [],
   max = 10,
   min = 1,
   onChange = () => {},
   onChangeCommitted = () => {},
   orientation = "horizontal",
-  predefValues = [],
-  labels = true,
-  labelPosition = "bottom",
+  showLabels = true,
   step = 1,
   title,
   value,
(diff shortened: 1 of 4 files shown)


src/Slider/Slider.js
```diff
@@ -0,0 +1,150 @@
...
+    </Box>
+  );
+}
+
+Slider.propTypes = {
+  /**
+   * If auto the value label will display when the thumb is hovered;
+   * if on, will diplay persistently and if off will never display
+   */
+  displayCurrentValue: PropTypes.oneOf(["auto", "off", "on"]),
```
dmbarry86, line 89: Rename this to match the MUI prop name so users don't have to learn another syntax. MUI prop name is valueLabelDisplay

src/Slider/Slider.stories.js
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <div style={{ width: 500 }}>
+      <Slider {...args} onChange={onChange} value={value} />
+    </div>
+  );
+};
+
+// default story
+export const Default = Template.bind({});
+Default.args = {
```
dmbarry86, line 31: MIssing some default prop values here which means Storybook loads with no selections.

image

src/Slider/Slider.test.js
```diff
@@ -0,0 +1,24 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import Slider from "./Slider";
+
+/**
+ * Tests
+ */
+describe("Slider", () => {
+  test("can renders without crashing", () => {
```
dmbarry86, line 9: This sounds a bit brutal 'without crashing'. Maybe something a bit more like 'can render with no props'.

src/Slider/Slider.stories.js
```diff
@@ -0,0 +1,54 @@
...
+  min: 1,
+  step: 1,
+  title: "Slide example"
+};
+
+// custom labels story
+export const CustomLabels = Template.bind({});
+CustomLabels.args = {
+  displayCurrentValue: "auto",
+  labels: true,
```
dmbarry86, line 44: This is confusing, especially with predefValues. I would expect labels to be the cusotm labels I pass in whereas it's actually a boolean to show/hiode labels. Thus I would rename labels to showLabels or displayLabels to make it clear it's a boolean. I would then rename predefValues to labels.

src/Slider/Slider.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const { container } = render(<Slider />);
+    expect(container.firstChild).toBeInTheDocument();
+  });
+  test("can set the title", () => {
+    const title = "This is a test";
+    render(<Slider title={title} />);
+    expect(screen.getByText(title));
+  });
+  test("can set min value", () => {
```
dmbarry86, line 18: Should also add a test for setting max value.

src/Slider/Slider.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const minValue = 5;
+    const { container } = render(<Slider min={minValue} />);
+    const input = container.querySelector("input");
```
dmbarry86, line 21: What exactly is this input element you are testing for?

src/Slider/Slider.js
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Tick label position respective to slide
+   */
+  labelPosition: PropTypes.oneOf(["bottom", "top"]),
```
dmbarry86, line 93: Spacing between text and bar looks inconsistent between bottom and top. Seems less space in top but maybe it's just my eyes (I didn't measure).

image

image

src/Slider/Slider.js
```diff
@@ -0,0 +1,150 @@
...
+      stepSize = null;
+    }
+  } else {
+    marks = false;
+  }
+  // return components
+  return (
+    <Box>
+      <Typography>{title}</Typography>
+      {labelPosition === "top" ? (
```
dmbarry86, line 48: Would be cleaner to remove this ternary here, just have one instance of MuiSlider and then handle the sx styling as a ternary based on label position. Otherwise you are just duplicating code and could forget to add prop to both instances if any changes are made in the future.

src/Slider/Slider.js
```diff
@@ -0,0 +1,150 @@
...
+   * _value_: The new value.
+   */
+  onChangeCommitted: PropTypes.func,
+  /**
+   * Slider orientation
+   */
+  orientation: PropTypes.oneOf(["horizontal", "vertical"]),
+
+  /**
+   * Indicates predeterminated values to wich the user can move the slider.
```
dmbarry86, line 134: typo 'wich'

src/Slider/Slider.js
```diff
@@ -0,0 +1,150 @@
...
+   * If auto the value label will display when the thumb is hovered;
+   * if on, will diplay persistently and if off will never display
+   */
+  displayCurrentValue: PropTypes.oneOf(["auto", "off", "on"]),
+  /**
+   * Tick label position respective to slide
+   */
+  labelPosition: PropTypes.oneOf(["bottom", "top"]),
+  /**
+   * If true, thick labels should be displayed
```
dmbarry86, line 95: I think you mean 'tick' not 'thick'

dmbarry86 (changes requested):

dmbarry86:
Few items for Storybook:
* Default story no longer reflects the default component behaviour (red text)
* Custom labels story also includes custom colouring. Move this to separate story so we have one story showing one thing.

dmbarry86:
Rather than exposing a specific font prop I think it would be better to just expose a labelStyle prop and pass this on as sx to the label component. This way you have full control over all props and don't need to keep coming back to the component to add new props when you want to change another piece of style.

dmbarry86:
The first two stories still have labelColor and labelFont props in Storybook. Apart from this, all good.

PR #265, opened 2021-11-15
SidebarItem styling props

dmbarry86:
Added two new props to SidebarItem to allow custom styling to be passed in to the icon and text components. There are two new stories to show how to use them.


evoinea (approved):

evoinea:
LGTM

PR #268, opened 2021-11-17
Feature/text and number fields

jegasriskantha:
Added new NumberField and TextField components 

Contributes IPG-Automotive-UK/instrument-designer/issues/60 
Contributes IPG-Automotive-UK/instrument-designer/issues/61

TextField:
image

NumberField:
image


dmbarry86 (commented):
LGTM. Just one query over the number field. Is it possible to add a prop to disable the stepper? I can think of certain usecases where it wouldn't really make sense to have a stepper. For example in NGD upload wizard we would use number field but it wouldn't really provide any value to have a stepper that increments a value by 1 so we might want to disable it. If its not easy to do then don't worry for now.

jegasriskantha:
> LGTM. Just one query over the number field. Is it possible to add a prop to disable the stepper? I can think of certain usecases where it wouldn't really make sense to have a stepper. For example in NGD upload wizard we would use number field but it wouldn't really provide any value to have a stepper that increments a value by 1 so we might want to disable it. If its not easy to do then don't worry for now.

Done. I have added the ability to disable stepper, I also exposed the inputProps argument as it also allows you to set custom min, max and step (amongst other things) - this is what I am doing in the color component.

dmbarry86 (approved):

PR #275, opened 2021-11-23
Radio buttons group

evoinea:
Contributes to  https://github.com/IPG-Automotive-UK/instrument-designer/issues/8

New radio buttons group component. 
image

### DIFF
--- a/src/RadioButtons/RadioButtons.js
+++ b/src/RadioButtons/RadioButtons.js
@@ -46,10 +46,6 @@ export default function RadioButtons({
 
 RadioButtons.propTypes = {
   /**
-   * The default value
-   */
-  defaultValue: PropTypes.string,
-  /**
    * If true, the radio buttons group will be disabled
    */
   disabled: PropTypes.bool,
--- a/src/RadioButtons/RadioButtons.js
+++ b/src/RadioButtons/RadioButtons.js
@@ -14,19 +14,16 @@ export default function RadioButtons({
   disabled = false,
   labelPlacement = "end",
   onChange = () => {},
+  radioGroupStyle = {},
   row = false,
   title = "",
   values = []
 }) {
   return (
-    <FormControl
-      sx={{ border: "2px solid black" }}
-      disabled={disabled}
-      component="fieldset"
-    >
+    <FormControl disabled={disabled} component="fieldset">
       <FormLabel component="legend">{title}</FormLabel>
       <RadioGroup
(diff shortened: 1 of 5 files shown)


src/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,92 @@
...
+   */
+  onChange: PropTypes.func,
+  /**
+   * The lables of the radio buttons group
+   */
+  options: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Custom style to apply to the radio buttons group
+   */
+  radioGroupStyle: PropTypes.object,
```
dmbarry86, line 75: Can we maybe shorten this to group style or maybe even just style? I'm not a big fan of multi-word props. If you think there will be other styles then call it groupStyle. If this is the only style then just call it style.

src/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * The size of the radio button
+   */
+  size: PropTypes.oneOf(["small", "medium"]),
+  /**
+   * The title of the radio buttons group
+   */
+  title: PropTypes.string,
+  /**
+   * Value of selected radion button
```
dmbarry86, line 89: Typo 'radion'

src/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,92 @@
+import * as React from "react";
+import FormControl from "@mui/material/FormControl";
```
dmbarry86, line 2: We tend to stick to importing all components from the same package on one line.

`import {FormControl, FormLabel, Radio, ...} from @mui/material`

src/RadioButtons/RadioButtons.js
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * The lables of the radio buttons group
+   */
+  options: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Custom style to apply to the radio buttons group
+   */
+  radioGroupStyle: PropTypes.object,
+  /**
+   * The layout of the buttons
```
dmbarry86, line 77: I would update this comment to indicate it's a boolean. If true the buttons will be displayed in a row otherwise in a column.

dmbarry86 (changes requested):
Few small improvement suggestions.

dmbarry86 (approved):

PR #281, opened 2021-11-25
Feature/honor min max in numberfield

jegasriskantha:
- updated Numberfield to error and show helper message when value is not within min or max ranges. When a value is typed that is not within the given range the onChange callback is not fired.
- also updated default widths for TextField and NumberField to be full width

### DIFF
--- a/src/NumberField/NumberField.js
+++ b/src/NumberField/NumberField.js
@@ -9,29 +9,21 @@ export default function NumberField({
   disabled = false,
   error = false,
   helperText,
-  inputProps = {},
   label,
   margin = "normal",
+  max,
+  min,
   onChange = () => {},
   placeholder,
   required = false,
   size = "medium",
+  step = 1,
   stepper = true,
   value,
   variant = "outlined"
 }) {
   const [valueError, setValueError] = React.useState(false);
   const [errorMessage, setErrorMessage] = React.useState("");
-
-  // get max and min values from inputProps
-  let max;
-  if (inputProps.max !== undefined) {
-    max = parseFloat(inputProps.max);
-  }
-  let min;
-  if (inputProps.min !== undefined) {
-    min = parseFloat(inputProps.min);
-  }
 
   // handleChange
   const handleChange = event => {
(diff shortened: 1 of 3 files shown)


src/NumberField/NumberField.js
```diff
@@ -20,20 +20,53 @@ export default function NumberField({
...
     // update event.target.value so that it is a number
-    const newEventValue = { target: { value: Number(event.target.value) } };
+    const newValue = Number(event.target.value);
+    const newEventValue = { target: { value: newValue } };
     const updatedEvent = { ...event, ...newEventValue };
+
+    // if the value is less than minimum, set error
+    if (newValue < min) {
+      setValueError(true);
+      setErrorMessage(`Value must be greater than ${min}`);
```
dmbarry86, line 46:
```suggestion
      setErrorMessage(`Value must be greater than or equal to ${min}`);
```

src/NumberField/NumberField.js
```diff
@@ -20,20 +20,53 @@ export default function NumberField({
...
+      return;
+    }
+
+    // if the value is greater than maximum, set error
+    if (newValue > max) {
+      setValueError(true);
+      setErrorMessage(`Value must be less than ${max}`);
```
dmbarry86, line 53:
```suggestion
      setErrorMessage(`Value must be less than or equal to ${max}`);
```

src/NumberField/NumberField.js
```diff
@@ -20,20 +20,53 @@ export default function NumberField({
   value,
   variant = "outlined"
 }) {
+  const [valueError, setValueError] = React.useState(false);
+  const [errorMessage, setErrorMessage] = React.useState("");
+
+  // get max and min values from inputProps
+  let max;
```
dmbarry86, line 27: It might be cleaner and more obvious from a documentation point of view that the feature exists if min, max and step were their own separate props rather than inputProps. Maybe also a prop to disable the stepper would be good too.

dmbarry86: Ah you already have a prop to disable the stepper. I would go all the way and do the same for min, max and step then.

dmbarry86 (changes requested):
See suggestions in comments.

dmbarry86:
If I set a custom helper text then you'r custom error message doesn't shown. Is this intended behaviour?

image


image

src/NumberField/NumberField.js
jegasriskantha: Should I remove inputProps as a prop in the process?

jegasriskantha:
> If I set a custom helper text then your custom error message doesn't show. Is this intended behaviour? I think it would be better to show the error message and then switch back to helper text when no error exists.
> 
> image
> 
> image

It was intended behaviour with the idea that an explicitly set helper message would take priority over the internal message, but  I'm happy to change it.

src/NumberField/NumberField.js
dmbarry86: I would unless you can think of a reason to keep it.

dmbarry86:
> It was intended behaviour

Fair enough, keep as is then.

dmbarry86 (approved):

PR #282, opened 2021-11-26
Fix vertical slider

evoinea:
* Fixed for vertical slider.
* Added more options for label position: "left" and "right" for vertical sliders.

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -44,14 +44,6 @@ export default function Slider({
   } else {
     marks = false;
   }
-  // prevent horizontal keyboard navigation for vertical slider
-  const preventHorizontalKeyboardNavigation = (event, orientation) => {
-    if (orientation === "vertical") {
-      if (event.key === "ArrowLeft" || event.key === "ArrowRight") {
-        event.preventDefault();
-      }
-    }
-  };
   // set label position based on orientation
   let style = {};
   if (orientation === "horizontal") {
@@ -92,7 +84,6 @@ export default function Slider({
           onChange={onChange}
           onChangeCommitted={onChangeCommitted}
           orientation={orientation}
-          onKeyDown={preventHorizontalKeyboardNavigation}
           step={stepSize}
           value={typeof value !== "undefined" ? value : null}
           valueLabelDisplay={valueLabelDisplay}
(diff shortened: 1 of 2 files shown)


src/Slider/Slider.stories.js
```diff
@@ -131,3 +141,21 @@ Disabled.args = {
   value: 1,
   valueLabelDisplay: "auto"
 };
+
+// vertical slider
+export const VericalSlider = Template.bind({});
```
jegasriskantha, line 146:
```suggestion
export const VerticalSlider = Template.bind({});
```

src/Slider/Slider.stories.js
```diff
@@ -131,3 +141,21 @@ Disabled.args = {
...
+VericalSlider.args = {
```
jegasriskantha, line 147:
```suggestion
VerticalSlider.args = {
```

src/Slider/Slider.js
```diff
@@ -44,31 +44,60 @@ export default function Slider({
   } else {
     marks = false;
   }
+  // prevent horizontal keyboard navigation for vertical slider
+  const preventHorizontalKeyboardNavigation = (event, orientation) => {
```
jegasriskantha, line 48: orientation is never actually passed into preventHorizontalKeyboardNavigation so orientation is always undefined and the keyboard controls are still active. 

I don't think orientation needs to be passed in here as  it doesn't change for the lifetime of the component.

src/Slider/Slider.js
```diff
@@ -44,31 +44,60 @@ export default function Slider({
...
+    if (orientation === "vertical") {
+      if (event.key === "ArrowLeft" || event.key === "ArrowRight") {
+        event.preventDefault();
+      }
+    }
+  };
+  // set label position based on orientation
+  let style = {};
+  if (orientation === "horizontal") {
+    style = {
```
jegasriskantha, line 58: If ArrowLeft and ArrowRight are disabled for a vertical slider should the up and down arrows be disabled for the horizontal slider?

jegasriskantha (changes requested):
I'm not sure this is working as expected, see suggestions.

jegasriskantha (approved):
LGTM

PR #290, opened 2021-12-01
Bugfix/numberfield

jegasriskantha:
closes: IPG-Automotive-UK/instrument-designer/issues/84
closes: IPG-Automotive-UK/instrument-designer/issues/87 found in this PR: IPG-Automotive-UK/instrument-designer/pull/81

This fix introduces new internal state for the value of the NumberField, this enables the NumberField to display  what the user has typed without updating the 'external' controlling state. If a value is entered by the user that is outside the defined bounds the onChange event is not fired.  The error states are now managed by a useEffect, which means an error will be rendered if the initial value is outside the defined bounds.

### DIFF
--- a/src/NumberField/NumberField.js
+++ b/src/NumberField/NumberField.js
@@ -31,7 +31,7 @@ export default function NumberField({
     setNumber(value);
   }, [value]);
 
-  // validate the number field when number, min or max changes
+  // validate the number field when, number, min or max changes
   useEffect(() => {
     if (min && number < min) {
       setValueError(true);
@@ -55,8 +55,20 @@ export default function NumberField({
     // set the number
     setNumber(event.target.value);
 
+    // is min satisfied
+    let minSatisfied = true;
+    if (min && newValue < min) {
+      minSatisfied = false;
+    }
+
+    // is max satisfied
+    let maxSatisfied = true;
+    if (max && newValue > max) {
+      maxSatisfied = false;
+    }
+
     // // if the value is valid, set error to false
-    if (newValue >= min && newValue <= max && event.target.value !== "") {
+    if (minSatisfied && maxSatisfied && event.target.value !== "") {
       setValueError(false);
       setErrorMessage("");
       onChange(updatedEvent);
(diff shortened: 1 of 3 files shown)


src/NumberField/NumberField.js
```diff
@@ -32,25 +52,34 @@ export default function NumberField({
...
-    // if the value is less than minimum, set error
-    if (newValue < min && min !== undefined) {
-      setValueError(true);
-      setErrorMessage(`Value must be greater than or equal to ${min}`);
-      return;
+    // set the number
+    setNumber(event.target.value);
+
+    // is min satisfied
+    let minSatisfied = true;
```
dmbarry86, line 59:
```suggestion
    let minSatisfied = min && newValue >= min;
```

src/NumberField/NumberField.js
```diff
@@ -32,25 +52,34 @@ export default function NumberField({
...
     }
 
-    // if the value is greater than maximum, set error
-    if (newValue > max && max !== undefined) {
-      setValueError(true);
-      setErrorMessage(`Value must be less than or equal to ${max}`);
-      return;
+    // is max satisfied
+    let maxSatisfied = true;
+    if (max && newValue > max) {
```
dmbarry86, line 66: Your logic here won't work when max value is set to 0 because 0 !== true so your if statement will always be skipped. You can test this by setting max to 0 and min to some negative number and then type in a value greater than 0 and you won't get error message. You should also check the same can't occur in your min logic on line 60.

src/NumberField/NumberField.js
```diff
@@ -32,25 +52,34 @@ export default function NumberField({
...
+      maxSatisfied = false;
+    }
+
+    // // if the value is valid, set error to false
```
dmbarry86, line 70: remove double //

dmbarry86:
Losing the error message when I click away is not what I would expect because the control is still displaying invalid number so error should remain. I don't really understand why you need to deal with onBlur here. Having the effect based on value should be enough.

dmbarry86 (changes requested):
Not working as expected.

dmbarry86 (approved):
All good now.

PR #298, opened 2021-12-11
Feature/numberfield error message prop

jegasriskantha:
Added a new props: 
- showMinMaxErrorMessage (default: true) 
False set here:
image

- startAdornment (default: empty string)
$ set here:
image

- endAdornment (default: empty string) 
px set here
image


dmbarry86 (approved):
LGTM. I'll leave you to merge and deal with release.

PR #304, opened 2021-12-15
Feature/expose colour in switchfield

jegasriskantha:
contributes to: IPG-Automotive-UK/instrument-designer/issues/121

This pull request: 
- adds new thumbColor and trackColor property to easily change the color properties of the MUI switchField 
- adds new switchStyle property to make it easier to customise the MUI switchField
- conditionally renders the FormHelperText only when the helperText prop is set. 

image

### DIFF
--- a/src/SwitchField/SwitchField.js
+++ b/src/SwitchField/SwitchField.js
@@ -19,9 +19,21 @@ export default function FormSwitch({
   helperText,
   label,
   onChange,
-  options
+  options,
+  switchStyle,
+  thumbColor,
+  trackColor
 }) {
   // return components
+
+  // color styling
+  const colorStyling = {
+    "& .MuiSwitch-thumb": { color: thumbColor },
+    ".Mui-checked": {
+      "& + .MuiSwitch-track": { backgroundColor: trackColor }
+    }
+  };
+
   return (
     <FormControl disabled={disabled}>
       {label && <FormLabel style={{ fontSize: "0.75rem" }}>{label}</FormLabel>}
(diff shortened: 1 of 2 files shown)


evoinea:
As discussed, I think it might be a good idea to expose text colours too

src/SwitchField/SwitchField.js
```diff
@@ -105,5 +123,17 @@ FormSwitch.propTypes = {
         `Invalid prop ${propName} supplied to ${componentName}. Expected a string array of length 2.`
       );
     }
-  }
+  },
+  /**
+   * Custom styling for the MUI Switch component.
+   */
+  switchStyle: PropTypes.string,
```
dmbarry86, line 130: What's the purpose of this switchStyle prop? What would we use it for? If we need something non-standard then we just override the theme in the app where we need to. Also, should it not be an object anyway not a string? I think just remove this.

dmbarry86:
The track color is not actually the color I specify. It's the color I specify with some opacity on top. This won't work how you think it will in ID (which presumably is what you have in mind).

For example here I set both to red but only the thumb is red.
image

To be honest I don't really see why we need to complicate the API with these color props as 99% of the time you would just want it to match the theme of the app, e.g. primary. For the instrument designer usecase we can just write a custom style override (like we do for button color for example).

dmbarry86:
> * conditionally renders the FormHelperText only when the helperText prop is set.

I don't see any code you have added for this? It looks like the code already does this anyway on line 64 which has been there since day 1 of the component. Am I missing something?

dmbarry86 (commented):
I'm not convinced about the need for this PR. Let's discuss tomorrow.

dmbarry86:
> As discussed, I think it might be a good idea to expose text colours too

I don't think we should do this. I don't want to start complicating the API of these components with non-standard styling props. If we don't want the default MUi styling in a particular app then we just override it.

dmbarry86:
See https://github.com/IPG-Automotive-UK/instrument-designer/pull/140

PR #305, opened 2021-12-15
Removed bottom border for last table row

evoinea:
Removed bottom border for the last table row; it has a bit of shadow effect.
image
Updated table:
image

### DIFF
--- a/src/MultiText/MultiText.js
+++ b/src/MultiText/MultiText.js
@@ -11,7 +11,7 @@ import React from "react";
 export default function LabelSetter({
   onChange = () => {},
   rows = [],
-  style = { height: 250, width: "100%" }
+  style = { height: 254, width: "100%" }
 }) {
   // add an id for each label/value
   const rowsWithID = rows.map((item, index) => ({ ...item, id: index }));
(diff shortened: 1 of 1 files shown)


dmbarry86:
I don't think this fix is a good solution to the problem you are trying to solve.

The missing border now looks strange and it's like the last row is not really a row. You've really only solved the problem for the very specific case where the bottom table row aligns exactly with the bottom of the table outer border.

image

image

src/MultiText/MultiText.js
```diff
@@ -11,7 +11,7 @@ import React from "react";
 export default function LabelSetter({
   onChange = () => {},
   rows = [],
-  style = { height: 250, width: "100%" }
+  style = { height: 254, width: "100%" }
```
dmbarry86, line 14: Why is this seemingly random 254 pixels the default style? In fact, why do I need to pass style at all?

dmbarry86 (changes requested):
See comments.

dmbarry86:
Not going ahead with this as just discussed on our call.

PR #307, opened 2021-12-15
Fixed the bug with the step value smaller than 1

evoinea:
Closes https://github.com/IPG-Automotive-UK/instrument-designer/issues/125.

Fixed the issue with the step size being < 1.
image


dmbarry86 (approved):
The fix works ok. In future though, please always add a new test when fixing a bug to ensure that it never pops up again. I have added one for you on this occassion.

PR #308, opened 2021-12-15
Fix close button size

dmbarry86:
Closes https://github.com/IPG-Automotive-UK/instrument-designer/issues/133

Fixed issue with the close button in snackbar showing as an elipse rather than circle when hovering over it. This was introduced during the recent migration to MUI 5 with updated styling. As well as make it circular again I have also knocked down the size a little bit as it was overwhelming the snackbar a bit too much.


jegasriskantha (approved):
LGTM

PR #310, opened 2021-12-16
Slider should not overflow its element width

evoinea:
Contributes to https://github.com/IPG-Automotive-UK/instrument-designer/issues/99.

Updated Slider to not overflow it's parent container.

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -68,9 +68,17 @@ export default function Slider({
   }
   // return components
   return (
-    <Box sx={{ height: "100%" }}>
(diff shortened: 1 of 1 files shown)


dmbarry86:
The distance between the slider and the text seems inconsistent left to right.

Compare here the distance between the circle and the number 4. Left seems to have more space between.
image


image

dmbarry86:
I think same is true for top compared to bottom although not quite as much. Seems here that text in top has larger gap compared to text on bottom.

image

dmbarry86 (commented):
Overflow issue seems to be fixed now. I did spot a potential incosistency in the text spacing though which maybe you could take a look at.

Issue #58, opened 2020-11-29
Add DialogTitle component


dmbarry86:
Migrate the DialogTitle component currently used in a number of our apps over to this library.

dmbarry86:
Test that onClose gets called and another test that childen (text) get rendered.

dmbarry86:
Story to create example dialog with dialog title as child.

dmbarry86:
Repos to update:
* NGD
* VPM
* MM
* Cloud licensing
* Vehicle manager (check with @jegasriskantha if any code has been written yet)

dmbarry86:
I would suggest to switch over from withStyles to makeStyles for consistency with other component styling. I can help when you get to this.

PR #378, opened 2022-04-29
Feedback form

nikitharani:
Closes https://sce.myjetbrains.com/youtrack/issue/TD-309

Added a new feedback form component to capture customer feedback inside our web apps.

The form is launched from a floating action button.
image

First user selects the feedback type.
image

Next they can enter title and detailed description and click submit. onSubmit callback prop is exposed to allow the componet to be integrated into our apps and values user entered are returned in the callback data.
image


src/FeedbackForm/FeedbackForm.test.js
```diff
@@ -0,0 +1,37 @@
...
+    userEvent.type(title, "Car Maker");
+
+    const description = screen.getByRole("textbox", { name: /description/i });
+    userEvent.type(description, "Car Maker");
+    userEvent.click(screen.getByRole("button", { name: /submit/i }));
+
+    await waitFor(() => {
+      expect(onSubmit).toHaveBeenLastCalledWith({
+        description: "Car Maker",
+        title: "Car Maker",
```
dmbarry86, line 30: It would be better to use different text for the title and description so we can guarantee they are correct and not mixed up in the code. For example, imagine you made mistake in the code and title was being returned as description, this test wouldn't catch it.

src/FeedbackForm/FeedbackForm.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  beforeEach(() => {
+    render(<FeedbackForm onSubmit={onSubmit} />);
+  });
+
+  it("onSubmit is called when all fields pass Validation", async () => {
+    const floatingButton = screen.getByRole("button", { name: /add/i });
+    userEvent.click(floatingButton);
+
+    const feedback = await screen.findByTestId("SentimentSatisfiedAltIcon");
```
dmbarry86, line 17: This test should be repeated for all 3 sentiment options to ensure they all work.

src/FeedbackForm/FeedbackForm.test.js
```diff
@@ -0,0 +1,37 @@
...
+    const title = await screen.findByRole("textbox", { name: /title/i });
+    userEvent.type(title, "Car Maker");
+
+    const description = screen.getByRole("textbox", { name: /description/i });
+    userEvent.type(description, "Car Maker");
+    userEvent.click(screen.getByRole("button", { name: /submit/i }));
+
+    await waitFor(() => {
+      expect(onSubmit).toHaveBeenLastCalledWith({
+        description: "Car Maker",
```
dmbarry86, line 29: I think the sentiment selection should also be expected in the onSubmit data object alongside description and title etc.

src/FeedbackForm/FeedbackForm.js
```diff
@@ -0,0 +1,268 @@
...
+  const [openFeedback, setOpenFeedback] = React.useState(false);
+
+  const handleClickOpen = () => {
+    setOpenFeedback(true);
+  };
+  const handleClose = () => {
+    setOpenFeedback(false);
+  };
+
+  // child dialog
```
dmbarry86, line 67: There are few places where comments could be improved in this file as it is quite long. I will make the changes for you so you can see an example for next time.

src/FeedbackForm/FeedbackForm.js
```diff
@@ -0,0 +1,268 @@
...
+import { useState } from "react";
+
+// for parent dialog
+const BootstrapDialog = styled(Dialog)(({ theme }) => ({
+  "& .MuiDialogContent-root": {
+    padding: theme.spacing(2)
+  }
+}));
+
+const BootstrapDialogTitle = props => {
```
dmbarry86, line 29: I think this dialog title component should be separate component inside the react-ui library. We already have a ticket open to migrate dialog title to new component so would be great if you could do that as part of this work. We use dialog title in few other apps and at the moment it is duplicated so would be nice to have it as separate component.

https://github.com/IPG-Automotive-UK/react-ui/issues/58

dmbarry86:
I think would look better if the sentiment feedback dialog is not shown behind the title and description dialog. I think also would be good if the height of the two dialogs was the same so user feels like it is just one dialog, which to be honest it could be.

image

dmbarry86:
Description should be multi-line text field. Maybe default to something like 3 rows.

image

dmbarry86:
The open prop doesn't seem to be working. I would expect dialog to open when I set to true. Please also add a new test for this.

image

dmbarry86:
i think would be good if the feedback dialog was anchored to the position of the button rather than in middle of screen like it as at the moment. Maybe also there needs to be prop to control which position it is anchored to. In the app suite we probably will have the floating button in bottom right of the screen and then I would want the feedback form to also appear in bottom right.

image

dmbarry86:
Rather than 'message' I think 'sentiment' is a better property name. Looks like the MUI icons already use sentiment also.

image

dmbarry86:
index.js file at top level should be updated to export your component so it can be imported by apps that use this library.

image

dmbarry86 (changes requested):
I've left a few suggestions for improvement. Just let me know if you have any questions.

dmbarry86:
Could you please rename BootstrapDialogTitle to just DialogTitle?

dmbarry86:
@nikitharani I've pushed updates for following changes we discussed earlier. Please check if they look OK to you.

1. Fixed the tests for dialog open/close.
2. Added parameterised test for the 3 different sentiment types.
3. Fixed issue with sentiment not being returned in onSubmit.
4. Re-factored the component to just contain one dialog rather than two.

dmbarry86:
Please rename to DialogTitle not Dialogtitle. We try to use CamelCase for multi-word names.

dmbarry86:
> I think would look better if the sentiment feedback dialog is not shown behind the title and description dialog. I think also would be good if the height of the two dialogs was the same so user feels like it is just one dialog, which to be honest it could be.
> 
> image

I made these changes in my last set of commits.

dmbarry86:
> > DialogTitle
> > renaming to DialogTitle has a conflict with built-in react DialogTitle component. so I am renaming it to DialogeTitle.

You can resolve this by aliaising the MUI DialogTitle when you import it.

import {DialogTitle as MuiDialogTitle} from '@mui/material'.

You will see I already do this for the Dialog itself when importing inside FeedbackForm.

src/index.js
```diff
@@ -3,7 +3,6 @@ export { default as Autocomplete } from "./Autocomplete";
 export { default as Checkbox } from "./Checkbox";
 export { default as Copyright } from "./Copyright";
 export { default as Color } from "./Color";
-export { default as DialogTitle } from "./DialogTitle";
```
dmbarry86: This should be added back. We want the DialogTitle to be a component that can be exported from this library and not just an internal component to the library.

dmbarry86:
Please can you add a story and test for the DialogTitle component?

The story would probably need to create a dialog so you can show the dialog title component.

I think 3 tests.
1. Check title is correctly shown.
2. Check the close button is shown when onClose is provided.
3. Check the close button is not shown when onClose is not provided.

dmbarry86:
Your new position property needs to be exposed as a property which user can change. Should also define a default.

Should appear here
image


and here
image

dmbarry86 (changes requested):
Few changes to make. See comments above.

dmbarry86 (approved):
Looks really good.

Issue #297, opened 2021-12-07
Do not fire onChange and show error/ helper message when value is not within the interval set by step size. 


jegasriskantha:
If a step size is set in NumberField that step size is used to specify the interval between legal numbers. Currently if an illegal number is set a hover over message is shown as seen below: 

![image](https://user-images.githubusercontent.com/75164177/145019446-e966a7d1-40ef-48c2-af85-afd7093469a7.png)

It would be ideal if the message showed as an error helper message and did not fire an onChange event

PR #401, opened 2022-06-07
Error when enter incorrect value

evoinea:
Fixes https://github.com/IPG-Automotive-UK/react-ui/issues/297.

Updated NumberField component to error if an incorrect number is set (not a multiple of step).
image

I also have fixed the existing warning regarding the changing of the controlles input to be uncontrolled.
 
image

### DIFF
--- a/src/NumberField/NumberField.js
+++ b/src/NumberField/NumberField.js
@@ -31,8 +31,13 @@ export default function NumberField({
 
   // update the value of the number field
   useEffect(() => {
+    console.log(value);
     setNumber(value);
   }, [value]);
+
+  const roundAccuratley = (number, decimalPlaces) => {
+    Number(Math.round(number + "e" + decimalPlaces) + "e-" + decimalPlaces);
+  };
 
   // validate the number field when, number, min or max changes
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


dmbarry86 (changes requested):
Error with values shown in error message.

dmbarry86:
Should also add some new tests for this new behaviour. Should test that error state and message are correctly shown (maybe use the values I used here as an example) and also other tests that onChange does not get called for invalid value but does for valid value.

src/NumberField/NumberField.js
```diff
@@ -34,19 +34,53 @@ export default function NumberField({
...
     } else if (max !== undefined && number > max) {
+      // value is greater than max
       setValueError(true);
       if (showMinMaxErrorMessage) {
-        setErrorMessage(`Must be less than or equal to ${max}`);
+        setErrorMessage(`Must be less than or equal to ${max}.`);
       }
+    } else if (
+      number !== undefined &&
+      ((number * 100) % (step * 100)) / 100 !== 0
```
dmbarry86, line 65: Why do you need to multiply each term by 100?

dmbarry86 (approved):
Issue with incorrect helper message is fixed now. Works fine.

src/NumberField/NumberField.js
evoinea: To avoid rounding error

PR #402, opened 2022-06-08
Switch button darkmode

nikitharani:
closes https://sce.myjetbrains.com/youtrack/issue/TD-173

A Switch to toggle between dark and light mode.


dmbarry86:
You are not exporting the component from the root index file so it won't be possible for other apps to import it.

src/SwitchMode/ToggleColorMode.test.js
```diff
@@ -0,0 +1,57 @@
...
+    render(<ColorModeWithState mode="dark" />);
+    const button = screen.getByRole("button");
+
+    fireEvent.click(button);
+    expect(screen.getByTestId("Brightness4Icon"));
+
+    fireEvent.click(button);
+    expect(document.querySelector("#sandbox > svg > path"));
+  });
+});
```
dmbarry86, line 57: These tests for value change should connect the onChange callback to a jest function and check the function has been called with the correct value rather than checking correct image is shown.

dmbarry86: Something like this.

```
// setup component
const onChange = jest.fn();
render(<Component onChange={onChange}/>);

// click the button

// check on change has been called
expect(onChange).toHaveBeenCalledWith(blah blah)
```

dmbarry86:
I think an actual switch rather than a button would look better here and is what you typically see.

Here's a few random examples I found on Google image search.

image

image

dmbarry86 (changes requested):
Few bits of feedback.

dmbarry86:
New component works fine. I did a local build and tested it in the toolbar of one of our apps.

dmbarry86 (approved):

PR #403, opened 2022-06-13
Create dark theme

nikitharani:
Closes https://sce.myjetbrains.com/youtrack/issue/TD-448

Added DarkTheme palette in ThemeProvider.

### DIFF
--- a/src/ThemeProvider/ThemeProvider.js
+++ b/src/ThemeProvider/ThemeProvider.js
@@ -5,7 +5,7 @@ import {
 import PropTypes from "prop-types";
 import React from "react";
 
-// custom material-ui theme
+// custom material-ui theme for light mode
 const LightTheme = createTheme({
   overrides: {
     MuiAccordionSummary: {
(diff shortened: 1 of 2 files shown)


src/ThemeProvider/ThemeProvider.js
```diff
@@ -51,11 +51,22 @@ const theme = createTheme({
...
+    mode: "dark"
+  }
+});
+
 /**
  * IPG Material-ui theme provider
  */
-export default function ThemeProvider({ children }) {
-  return <MuiThemeProvider theme={theme}>{children}</MuiThemeProvider>;
+export default function ThemeProvider({ children, mode }) {
```
dmbarry86, line 64: set default to light so that if ThemeProvider is called without this new mode prop then we still get same behaviour as today with default being light.

```suggestion
export default function ThemeProvider({ children, mode = "light" }) {
```

src/UserMenu/UserMenu.js
```diff
@@ -125,7 +125,8 @@ const UserAvatar = ({ username, ...rest }) => {
       {...rest}
       sx={{
         background: "none",
-        border: theme => `2px solid ${theme.palette.background.paper}`,
+        border: "2px solid white",
```
dmbarry86, line 128: I guess this fixed the user avatar in both dark and light mode? Could you attach screenshot of each version just to confirm and to save me from having to manually build?

dmbarry86 (changes requested):
Keep same default light mode.

src/UserMenu/UserMenu.js
nikitharani: Yes, changing the border-color resolved the issue.
dark mode:
image
lightmode:
image

Issue #392, opened 2022-05-17
Migrate to React18


evoinea:
Migrate to React18 and fix all component errors (as a result of migration).

PR #405, opened 2022-06-24
Migrate to React18

evoinea:
Closes https://github.com/IPG-Automotive-UK/react-ui/issues/392.

Updated react-ui to React18.


src/FeedbackForm/FeedbackForm.js
```diff
@@ -150,7 +150,7 @@ export default function FeedbackForm({
                     <span>I like Something</span>
                   </Typography>
                   <Typography sx={{ mb: 1 }} color="text.primary">
-                    We like to hear what we're doing right.
+                    {"We like to hear what we're doing right."}
```
dmbarry86, line 153: Why did you make this change? I don't see anything wrong with the old version. I see the change made in a few places, not just here.

src/Sidebar/SidebarItem/SidebarItem.stories.js
```diff
@@ -54,6 +54,7 @@ Nested.args = {
     <SidebarItem
       key="withCount"
       {...WithCount.args}
+      // eslint-disable-next-line react/no-children-prop
```
dmbarry86, line 57: Rather than disable this warning, can you not update the code? Rather than passing children prop it should be possible to just explicitly create them by nesting.

.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm install
+      - run: npm install -force
```
dmbarry86, line 25: Is this definitely needed? I didn't need to use force when installing locally. FIne if it is needed, just curious why I didn't need to run it.

.storybook/preview-head.html
```diff
@@ -1 +0,0 @@
-<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
```
dmbarry86, line 1: How come this was removed? Roboto is the default font we use in most apps so nice to show it in storybook too if we can.

.storybook/preview.js
```diff
@@ -1,9 +1,9 @@
-import React from "react";
-import { addDecorator } from "@storybook/react";
-import ThemeProvider from "../src/ThemeProvider/ThemeProvider";
-
-addDecorator(story => <ThemeProvider>{story()}</ThemeProvider>);
```
dmbarry86, line 5: We have lost the theme provider which means all of the components in storybook are showing up in default MUI colors rather than the IPG dark blue etc.

dmbarry86 (changes requested):
Few minor issues to look at and couple of questions.

dmbarry86:
I think you mentioned this was a known issue? If so, could you please link the ticket so we can keep track?

image

dmbarry86:
I guess this one has always been there too but maybe you could take a quick look. Is it related to the test or something in the component itself that needs updating?

image

src/FeedbackForm/FeedbackForm.js
evoinea: To fix eslint error regarding special characters: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md

src/PasswordChangeDialog/PasswordChangeDialog.js
```diff
@@ -229,7 +228,7 @@ const dialogTitleStyles = {
 const DialogTitle = ({ children, onClose, ...other }) => {
   return (
     <MuiDialogTitle sx={dialogTitleStyles.root} {...other}>
-      <Typography variant="h6">{children}</Typography>
+      <div variant="h6">{children}</div>
```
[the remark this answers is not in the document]
evoinea: By accident sorry. I've reverted it back.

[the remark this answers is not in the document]
evoinea: It was a fix for this warning:
image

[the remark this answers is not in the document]
evoinea: I found a way to fix this warning and to keep <Typography> too.

src/FeedbackForm/FeedbackForm.js
evoinea: As discussed, I disabled this eslint check and reverted the code.

.github/workflows/test.yml
evoinea: It is nedded, @use-debouncy library needs to be migrated to React18.

PR #407, opened 2022-07-01
TD-518 - Add "Back to App Store" Feature

chhack94:
This adds a Link element to the IPG Logo in the Top left Corner. The Link gets generated from a environment variable called "REACT_UI_APPSTORE_URL" in each app during the build event, if it is empty the behaviour stays the same as before this feature implementation.

Edit: using a prop with default = "#" for the Sidebar Item


dmbarry86:
The implementation here is not quite right. The URL should just be an additional prop on the Sidebar component which gets passed in from each app, not an environment variable. This UI library is completely independent of our app store so should know nothing about it. The prop should just be a generic prop. Each app then manages the actual app store URL during build time using the environment variable and passes it in as the prop value.

src/Sidebar/Sidebar.js
```diff
@@ -1,4 +1,5 @@
 import { Box, Divider, Typography } from "@mui/material";
+import Link from "@mui/material/Link";
```
dmbarry86, line 2: Should be imported on line 1 along with the other material component

dmbarry86:
Please add a new test for this new behaviour. I would expect two tests. The first that no link is shown when the prop is empty. The 2nd that the link is shown when prop is set and has correct URL.

dmbarry86 (changes requested):
Few things to update.

dmbarry86:
> Storybook would also need updating to include the new prop and add a new story for the link behaviour.

Don't forget this.

dmbarry86 (changes requested):
Add new story to storybook for Sidebar component.

chhack94:
> > Storybook would also need updating to include the new prop and add a new story for the link behaviour.
> 
> Don't forget this.

my bad, sorry

dmbarry86:
New story is good but all of the other stories against sidebar now also have a link which when clicked takes to a page with just the sidebar which isn't quite right. I think all other stories shouldn't be showing link at all. Looks like there is a # in there somewhere causing the issue. I think default should just be empty.

dmbarry86 (changes requested):
Link is still being shown even when logoLinkUrl prop is not set.

dmbarry86 (approved):

PR #412, opened 2022-07-12
Searchbar and scrollbar dark mode

nikitharani:
The border for the Search bar and color for the Scroll bar will be changed based on the mode

### DIFF
--- a/src/SearchBar/SearchBar.js
+++ b/src/SearchBar/SearchBar.js
@@ -16,7 +16,10 @@ export default function SearchBar({
       variant="outlined"
       sx={{
         alignItems: "center",
-        borderColor: "rgba(0, 0, 0, 0.23)",
+        borderColor: theme =>
+          theme.palette.mode === "light"
+            ? "rgba(0, 0, 0, 0.23)"
+            : "theme => theme.palette.text.secondary",
         display: "flex",
         marginBottom: theme => theme.spacing(2),
         marginTop: theme => theme.spacing(2)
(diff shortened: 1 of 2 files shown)


src/SearchBar/SearchBar.js
```diff
@@ -16,7 +16,10 @@ export default function SearchBar({
       variant="outlined"
       sx={{
         alignItems: "center",
-        borderColor: "rgba(0, 0, 0, 0.23)",
+        borderColor: theme =>
+          theme.palette.mode === "light"
+            ? "rgba(0, 0, 0, 0.23)"
```
dmbarry86, line 21: Do we need to hardcode color here? Can we not reference theme? Is theme.palette.divider not the default that we want to keep in light mode?

dmbarry86: To be honest, I think you could just use theme.palette.divider for both dark and light mode. I imagine this is the default anyway so maybe you can just remove borderColor completely.

I also think your "theme => theme.palette.text.secondary" is not doing what you think. You have put it in a string so it's not actually evaluating. I think theme.palette.text.secondary is actually white in dark mode so what's happening is it's just reverting to default anyway because it doesn't recognise your string. If you did want to reference theme then it shouldn't be inside a string and should just be theme.palette.text.secondary. Anyway, I think removing borderColor and relying on defaults will work for both dark and light mode.

dmbarry86 (changes requested):
See comment.

dmbarry86 (approved):

PR #414, opened 2022-07-25
Transfer list component

nikitharani:
Transfer list component


dmbarry86:
Component doesn't look right in storybook. Seems to be missing all the lines and search box etc.

Storybook.
image

Vs the prototype I created for you.
image

src/TransferList/TransferList.js
```diff
@@ -0,0 +1,228 @@
...
+  Checkbox,
+  IconButton,
+  List,
+  ListItem,
+  ListItemIcon,
+  ListItemText,
+  Typography
+} from "@mui/material";
+import { React, useEffect, useState } from "react";
+import Box from "@mui/material/Box";
```
dmbarry86, line 12: This should be imported with all of the other mui/material imports above.

src/TransferList/TransferList.js
```diff
@@ -0,0 +1,228 @@
...
+import CloseIcon from "@mui/icons-material/Close";
+import InputBase from "@mui/material/InputBase";
```
dmbarry86, line 14: This should be imported with all of the other mui/material imports above.

dmbarry86:
> Component doesn't look right in storybook. Seems to be missing all the lines and search box etc.
> 
> Storybook. image
> 
> Vs the prototype I created for you. image

Seems to be because theme variable was set to dark in application data which doesn't really work too well because of the white background. When in light mode it works ok.

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+describe("TransferList", () => {
+  test("Search bar updates value", async () => {
+    const user = userEvent.setup();
+    const { container } = render(<SelectedItemsWithState />);
+
+    const inputBase = container.querySelector(".MuiInputBase-input");
+    await user.type(inputBase, "pears");
+    expect(inputBase.value).toBe("pears");
```
dmbarry86, line 38: This isn't really checking the correct thing. All this test is doing is typing in search box and then checking that value in search box is correct, which of course it will be because user.type deals with that. We don't need to test user.type functionality. What we should be testing is that when search text is typed then the list on the left is filtered and shows the correct items.

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+    const user = userEvent.setup();
+    const { container } = render(<SelectedItemsWithState />);
+
+    const inputBase = container.querySelector(".MuiInputBase-input");
+    await user.type(inputBase, "p");
+
+    const list = screen.getByRole("list");
+    const { getAllByRole } = within(list);
+    const items = getAllByRole("listitem1");
+    expect(items.length).toBe(2);
```
dmbarry86, line 65: Would be good to check the actual items are correct rather than just how many there are, i.e. check that the text of each item is as expected, which I think in this case should be Apples and Pears but not Oranges.

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  test("test selected item in the list", async () => {
+    const user = userEvent.setup();
+    const onChange = jest.fn();
+
+    render(
+      <SelectedItemsWithState onChange={onChange} selectedItem={["Pears"]} />
```
dmbarry86, line 73: I don't think Pears should be the initially selectedItem because the whole point of this test is to select the item by clicking on it. I think selectedItem should be empty and then your test selects Pears item and checks callback is called with correct item.

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+    const user = userEvent.setup();
+
+    const onChange = jest.fn();
+    const { container } = render(
+      <SelectedItemsWithState onChange={onChange} selectedItem={[]} />
+    );
+
+    user.click(screen.queryByTestId("close"));
+
+    expect(container.querySelector(".MuiTypography-body1").textContent).toBe(
```
dmbarry86, line 94: I think it would be good to also check onChange gets called with correct item (empty in this case).

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+  test("On click clearAll button should has text 'None Selected'", () => {
+    const user = userEvent.setup();
+    const onChange = jest.fn();
+    const { container } = render(
+      <SelectedItemsWithState onChange={onChange} selectedItem={[]} />
+    );
+
+    user.click(screen.queryByTestId("clear-all"));
+
+    expect(container.querySelector(".MuiTypography-body1").textContent).toBe(
```
dmbarry86, line 50: I think it would be good to also check onChange gets called with correct item (empty in this case).

src/TransferList/TransferList.js
```diff
@@ -0,0 +1,228 @@
...
+  ),
+  /**
+   * Callback fired when the new item is selected.
+   *
+   * **Signature**
+   * ```
+   * function(newItems) => void
+   * ```
+   *
+   * _newItems_: The new item that has been selected
```
dmbarry86, line 219: Callback contains all items selected, not just the new item that was clicked.

```suggestion
   * _newItems_: The new items that have been selected
```

src/TransferList/TransferList.test.js
```diff
@@ -0,0 +1,98 @@
...
+      "None Selected"
+    );
+  });
+
+  test("on search sort the list items", async () => {
+    const user = userEvent.setup();
```
dmbarry86, line 56: Would be good to add a few comments to each test. Probably everywhere you have added new line there should be small comment before that section.

```
// comment
code...
...

// comment
code...
...
```

dmbarry86 (changes requested):
Few small suggestions to improve the tests.

dmbarry86 (changes requested):
As discussed on chat:
* Fix heights
* Fix items property in story

dmbarry86:
The height of the top component needs to be increased slightly. If you hover over the clear all button you will see it now overflows its container. I'll update for you.

dmbarry86 (approved):

PR #440, opened 2022-09-14
Bugfix/numberfield does not allow 0

jegasriskantha:
Summary:
- updated the NumberField component to allow an initial value of 0 
- updated the NumberField component to ignore step logic when no step is passed into the component 
- added tests for the above two cases.


mattcorner (approved):
LGTM

PR #442, opened 2022-09-20
Feature/numberfield refactor

mattcorner:
Refactored NumberField to reduce the complexity but maintain functionality. Hopefully this makes it a little easier to maintain.

* Reduced the amount of internals state we need to handle. We only care about the value.
* Base error state and error text off of existing state. We should be running this on every render so no need for additional state here.
* Store all validation logic in a single function to reduce complexity and improve code readability.

Should resolve an issue Dave found where value could be null.

MicrosoftTeams-image

### DIFF
--- a/src/NumberField/NumberField.js
+++ b/src/NumberField/NumberField.js
@@ -1,177 +1,99 @@
-import { InputAdornment, TextField as MuiTextField } from "@mui/material";
-import React, { useEffect } from "react";
+import { InputAdornment, TextField } from "@mui/material";
+import React, { useEffect, useState } from "react";
 import PropTypes from "prop-types";
 
 /**
- * NumberField components are used for collecting user provided information as a Number.
+ * Number fields let users enter and edit numbers.
  */
-export default function NumberField({
-  disabled = false,
-  endAdornment = null,
-  error = false,
-  helperText,
-  label,
-  margin = "normal",
-  max,
-  min,
-  onChange = () => {},
-  placeholder,
-  required = false,
-  showMinMaxErrorMessage = true,
-  size = "medium",
-  startAdornment = null,
-  step,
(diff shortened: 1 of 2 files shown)


evoinea:
I think something is nor correct when changing step size and then use the stepper: for example if I set step=2 and then change the value using the stepper, next value will be 3, with an error ( 1 was expecting value to be 4):
image

mattcorner:
> I think something is nor correct when changing step size and then use the stepper: for example if I set step=2 and then change the value using the stepper, next value will be 3, with an error ( 1 was expecting value to be 4): image

To be honest, I'm not sure where the step validation requirement comes from. We have no ability to know whether a value change came from a user typing, or the spinner. Suggest we raise a new issue to fully consider the step functionality. @evoinea

evoinea:
Looks good to me.

evoinea (approved):

dmbarry86:
I still get the error when passing null. This used to work before all of the recent number field changes.
image

If I pass empty string instead of null, as per first error message hint, then I get a different error. Appears that the component is not behaving correctly as a controlled component, which is how it's being used in our apps.
image

There's also a new error related to sx.
image

dmbarry86:
> Should resolve an issue Dave found where value could be null.

Does not appear to be the case. Just have a refactored version of the old component with the same issue plus a new one.

dmbarry86:
I've created a sandbox to help you.

Here's the old behaviour as of 2.20.0 and before. No errors in console. Accepts null as 'no value'.
https://codesandbox.io/s/numberfield-old-7zcl7i

Here's the same code but with 2.20.3. Errors in console.
https://codesandbox.io/s/numberfield-new-g72r2l?file=/src/App.js

PR #443, opened 2022-09-22
Bugfix/numberfield null

mattcorner:
* Fixed null being allowed as a value
* Fixed controlled vs uncontrolled
* Fixed floating point issue
* Fixed sx error


dmbarry86 (approved):
Looks good. Fingers crossed.

PR #458, opened 2022-10-05
Remove popover from color component

dmbarry86:
Removed the popover dialog and swatch from the color component. Color component is now just a standalone color selector component that can be parented wherever required. If a particular app requires it in a popover it can parent it so. We have a usecase in ID that doesn't require the color component with swatch and popover and instead we just want the core color selector parts. This has also significantly simplified the component with less props and less code to deal with popover related stuff.

Multi-color also required updating as it uses the color component internally.

image

### DIFF
--- a/src/Color/Color.js
+++ b/src/Color/Color.js
@@ -129,123 +129,117 @@ export default function Color({
 
   // define components
   return (
-    <Box>
-      <Box sx={{ display: "flex", flexDirection: "column" }}>
-        <Box>
-          {showPicker && (
-            <Box>
-              <Box sx={sx.colorPicker}>
-                {noColorChecked ? (
-                  <Box
-                    sx={{
-                      border: "1px solid #bdbdbd",
-                      borderRadius: "4px",
-                      height: "238px"
-                    }}
-                  />
-                ) : (
-                  <RgbaColorPicker
-                    color={rgbaObj}
-                    onChange={handleColorPickerChange}
-                    id="colorPicker"
-                  />
-                )}
-              </Box>
-            </Box>
-          )}
-          {showControls && (
-            <div>
-              <Checkbox
-                checked={noColorChecked}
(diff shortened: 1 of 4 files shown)


src/Color/Color.js
```diff
@@ -383,40 +261,6 @@ Color.propTypes = {
...
-   */
-  popoverPositionTop: PropTypes.number,
-  /**
-   * This is the width of the popover, this also effects the
-   * popover's children the colorpicker and controls.
-   * @default 250px
-   */
-  popoverWidth: PropTypes.string,
   /**
    * This determines if the RGB and HEX controls are shown.
```
mattcorner, line 265:
```suggestion
   * This determines if the RGB controls are shown.
```
We don't show a HEX input ever.

src/Color/Color.js
```diff
@@ -157,222 +127,130 @@ export default function Color({
...
-                    disabled={noColorChecked}
-                    id="red"
-                    type="number"
-                    size="small"
-                    margin="dense"
-                    max="255"
-                    label="Red"
-                    error={rgbaObj.r > 255}
-                    value={noColorChecked ? "" : rgbaObj.r}
+      <Box sx={{ display: "flex", flexDirection: "column" }}>
```
mattcorner, line 133: Not sure that we need three nested divs here. 2 of them seem to do nothing.

src/Color/Color.js
```diff
@@ -157,222 +127,130 @@ export default function Color({
...
+        <Box>
+          {showPicker && (
+            <Box>
```
mattcorner, line 136: Another nested div that seems to do nothing

mattcorner (changes requested):

mattcorner (approved):
Co-wrote fixed with @dmbarry86. Approved.

PR #460, opened 2022-10-07
added the ability to set size on autocomplete component

jegasriskantha:
##Summary:
- Allows size to be set on autocomplete component


m4manjesh:
Do you think both default and value in control should be the same?
Screenshot (86)

jegasriskantha:
> Do you think both default and value in control should be the same? Screenshot (86)

Done:
image

PR #467, opened 2022-10-18
Feature/status labels

mattcorner:
- Added new **StatusLabel** component for use in TestCenter.
- Added new **StatusMessage** and **StatusIcon** components which are combined to make us StatusLabel. These are exposed as they may be required for future usage in TestCenter.

Figma designs:
As a label:
image
Icon only:
image

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -4,7 +4,8 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
-    "@storybook/preset-create-react-app"
+    "@storybook/preset-create-react-app",
+    "storybook-dark-mode"
   ],
   framework: "@storybook/react",
   core: {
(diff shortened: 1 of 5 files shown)


dmbarry86:
I have some time so I will review this now.

mattcorner:
> Code looks good. Just one question. How do the status icons look in dark mode?

Looks fine to me, was this a question or a hint that they don't?

I've opened up another PR that adds dark mode to storybook so we can test this kind of thing more easily.

dmbarry86:
> Looks fine to me, was this a question or a hint that they don't?

Yes it was a question. Can you share screenshot of how they look?

mattcorner:
> 

image
image
image
image
image
image
image

Just maybe the the "running" needs to be a bit brighter?

dmbarry86:
> Just maybe the the "running" needs to be a bit brighter?

Looks OK I think. We can go with it.

PR #469, opened 2022-10-24
Feature/label selector updates

jegasriskantha:
Updates to label selector component: 

-  added new max label length prop and functionality
-  updated to use latest react-best-gradient-color-picker
-  removed internal label state 
- changed "save" button when create a new label to "add"
- add tests for max label length


src/LabelSelector/LabelSelector.js
```diff
@@ -314,6 +315,12 @@ LabelSelector.propTypes = {
    * @see https://material-ui.com/components/autocomplete/#limit-tags
    */
   limitTags: PropTypes.number,
+  /**
+   * The maximum length of a label name.
+   * @default 50
+   * @type {number}
+   */
+  maxLabelLength: PropTypes.number,
```
dmbarry86, line 323: Does this prop just relate to name or description too? I think the prop name `maxLabelLength` is ambigious. If it applies to name, which I suspect it does, then maybe `nameMaxLength` might be clearer and future proofs if we ever want to limit description field too.

dmbarry86:
The color picker is still quite laggy when creating a new creating. If you compare it to the Color component in React-UI you will see the difference and how smooth the Color component is compared to that used in the label. How come the label component is not just using the color component from react-ui anyway? Could it easily do so? Might solve the lagging problem and one less dependency.

dmbarry86:
Label Color title doesn't look the same as name and description title in dark mode. Also not left aligned the same as the name and description.

image

dmbarry86:
LabelSelector component shows sx prop in Storybook but I don't think this prop actually exists on the component. We shouldn't be exposing sx prop anyway on components and instead should expose specific props for specific things.

image

dmbarry86:
Default story should show the default value, not empty. Default is 50 I think?

image

dmbarry86:
Try to keep the props in alphabetical order in the doc if possible. Makes it easier to read through and find a given prop.

image

dmbarry86 (changes requested):
Functionality looks good but I've left a few suggestions for improvement.

jegasriskantha:
> The color picker is still quite laggy when creating a new label. If you compare it to the Color component in React-UI you will see the difference and how smooth the Color component is compared to that used in the label. How come the label component is not just using the color component from react-ui anyway? Could it easily do so? Might solve the lagging problem and one less dependency.

I thought we were going in the direction of the "react-best-gradient-color-picker", but I prefer the react-ui one so I have implemented that. I have had to make  a small change to the color component to have a showNoColor prop which is now included in this PR.

dmbarry86 (approved):

PR #470, opened 2022-10-25
Feature/filters

mattcorner:
Added Filter components to react-ui

image

Suggested testing order FilterStack > FilterPopper > CheckboFilter | LabelFilter | RangeFilter to best understand the use cases.

### DIFF
--- a/src/Filter/CheckboxFilter/CheckboxFilter.js
+++ b/src/Filter/CheckboxFilter/CheckboxFilter.js
@@ -1,0 +1,150 @@
+import * as React from "react";
+
+import {
+  Autocomplete,
+  Box,
+  Checkbox,
+  TextField,
+  autocompleteClasses
+} from "@mui/material";
+import { CheckBox, CheckBoxOutlineBlank } from "@mui/icons-material";
+
+import OpenAutocomplete from "../OpenAutocomplete";
+import PropTypes from "prop-types";
+
+/**
+ * A checkbox filter allows the user to select multiple options from a list.
+ */
+export default function CheckboxFilter({
+  variant = "popper",
+  limitTags = -1,
+  value = [],
+  ...props
+}) {
+  const defaults = { limitTags, value };
+  return variant === "popper" ? (
+    <CheckboxFilterInput {...props} {...defaults} />
+  ) : (
+    <CheckboxFilterInline {...props} {...defaults} />
+  );
+}
+
+/**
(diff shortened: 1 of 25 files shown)


jegasriskantha:
image

It would be could to keep the props in alphabetically order in the docs. They don't seem to be in order on a few of the components.

jegasriskantha:
Default story should show the default values for all props, not empty. (-1 here I believe)

image

jegasriskantha:
When the checkbox filter is set to always open and there is a long list of options. When I select an option lower down in the list (after scrolling) the options jump to the top of the list. I'm not sure if this is by design or easy to fix if not, just seems a bit weird to me.  
https://user-images.githubusercontent.com/75164177/197994401-8c5f5d0d-dcb8-4184-8afc-53a91b3e2f20.mp4

src/Filter/LabelFilter/LabelFilter.js
```diff
@@ -0,0 +1,193 @@
...
+   * { _id: 1, color: "#005FA8", description: "first label", name: "label 1" },
+   * ]
+   * @default []
+   */
+  value: PropTypes.arrayOf(PropTypes.object),
+  /**
+   * The variant of the filter
+   * @type {string}
+   * @default "popper"
+   * @enum {"popper", "inline"}
```
jegasriskantha, line 190: shouldn't the enums here be popper and always-open?

src/Filter/CheckboxFilter/CheckboxFilter.js
```diff
@@ -0,0 +1,151 @@
...
+   * @example
+   * ["Option 1"]
+   * @default []
+   */
+  value: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * The variant of the filter
+   * @type {string}
+   * @default "popper"
+   * @enum {"popper", "inline"}
```
jegasriskantha, line 148: shouldn't the enums here be popper and always-open?

jegasriskantha (changes requested):
Code looks good, just some minor comments.

mattcorner:
image

Issue spotted by @dmbarry86 on test-center. Proposed fix is that the popper always gets the same z-index as theme.zIndex.tooltip. Tested in test-center and works so implement here.

jegasriskantha (approved):
Nice, LGTM. I'll leave with you to merge & release.

PR #471, opened 2022-10-31
Feature/summary cards

jegasriskantha:
closes https://sce.myjetbrains.com/youtrack/issue/TD-889/Create-new-Metadata-card-in-react-ui

### Summary: 
- Implements new detail card component to view detailed metadata and hero image
- Implements new file card component to search and download files 
- Implements new summary card component to view summary metadata and hero image

### Detail Card
image

### File Card
image

### Summary Card
image


dmbarry86:
I think the image might look better left aligned rather than centered. All other items are left aligned (apart from edit/delete which are purposely on the right).

image

dmbarry86:
Also seems a bit weird that the full screen view shows image with different aspect ratio to the compact view. @Sowbhagya-ipg

dmbarry86:
Also seems like the detail card has fixed edit and delete buttons but compact view allows me to pass in my own options. What happens to my other options in full screen view? For example in VVB we have edit/delete/reconfigure. @Sowbhagya-ipg

dmbarry86:
I wonder if it would look better with the first column having a width based on the contents (up to some max) so in this example here the second column would be shifted left to align with the heading column and then all of the white space would be on the right? @Sowbhagya-ipg 

image

Sowbhagya-ipg:
> I wonder if it would look better with the first column having a width based on the contents (up to some max) so in this example here the second column would be shifted left to align with the heading column and then all of the white space would be on the right? @Sowbhagya-ipg
> 
> image

Yes, That might look nice. I will try that and see.

Sowbhagya-ipg:
> Also seems like the detail card has fixed edit and delete buttons but compact view allows me to pass in my own options. What happens to my other options in full screen view? For example in VVB we have edit/delete/reconfigure. Maybe I should just be able to pass in these options in details view as well and if I want delete/edit/reconfigure I can pass. @Sowbhagya-ipg

The idea is to have all the possible buttons at the top. If it includes reconfigure, then yes we should have reconfigure button at the top

Sowbhagya-ipg:
> Also seems a bit weird that the full screen view shows image with different aspect ratio to the compact view. @Sowbhagya-ipg

I thought of zooming in the image, but now i realise the images will pixelate if we zoom it that bigger. I will rethink the design for this full page view @dmbarry86

dmbarry86:
@Sowbhagya-ipg and @jegasriskantha what's the latest with the design refinements? Are we sticking how they are now or are you planning to implement some updates based on above comments? I'll continue reviewing the code implementation once I get the green light from you.

Sowbhagya-ipg:
@dmbarry86  I am working on the designs based on your feedback.

jegasriskantha:
@dmbarry86 still some work to do here, I have removed you as reviewer to stop you getting spammed with updates.  Will add you again when it is ready for review

dmbarry86:
@nikitharani the failing tests should be fixed before I start reviewing this.

nikitharani:
ready to review. @dmbarry86

dmbarry86:
It's probably a bit too late in the day now but for future referfence I think it's better to use component composition rather than exposing lots of different props for each possible customisable bit of the UI (content, buttonsStack etc). 

So we would follow something like Dialog in MUI or even Sidebar in this library.

Instead of:
```
<Card content={<div>Your Stuff </div>} buttonsStack={<div>Your Stuff </div>}>
```

Something like this
```
<Card>
   <CardContent>
      <div>Your Stuff</div>
   </CardContent>
   <CardActions>
      <div> Your Stuff </div>
   </CardActions>
</Card>
```

dmbarry86:
You have stories for DetailCard, SummaryCard and TableCard

image


But the library actually exports DetailCard, SummaryCard and FileCard.

image

dmbarry86:
Maybe I'm missing something but I don't see any action getting fired if I click on one of the file chips or the download all files button.

image

It looks like I pass an array of files rather than the UI elements themselves so I would expect some onClickFile, onClickDownloadAll etc to be exposed so use in each app.

dmbarry86:
Cards look a bit different in dark and light mode.

In light mode all of the backgrounds are white.

image

In dark mode some backgrouns are black and some are grey.
image

Is this intended behaviour?

dmbarry86 (changes requested):
- [ ] Evaluate effort to switch to component composition and feedback so we can make decision how to proceed.
- [x] Align library exports with stories.
- [x] Confirm dark/light mode is working as intended.
- [ ] Check file chip and file download button callbacks are exposed.

dmbarry86 (approved):
@nikitharani I'll leave you to deal with merging, updating release notes and making the release.

PR #503, opened 2023-01-18
Feature/image uploader

m4manjesh:
Closes https://sce.myjetbrains.com/youtrack/issue/TD-1190/Image-Uploader-react-ui-PT2

Figma: https://www.figma.com/proto/B75H0DFTEiUFjJtehIKa8M/Model-Manager?node-id=522%3A57530&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=228%3A28915


dmbarry86:
Is the dark mode implementationm how you would expect with the white background? Copying @Sowbhagya-ipg 

Dark mode
image

Light mode
image

dmbarry86:
Same comment regarding dark/light mode also applies for the file uplaoded component which was previously merged.

Dark
image


Light
image

dmbarry86:
The doc suggests there is only onAdd callback. Is there not also onDelete or maybe even just onChange that handles both? Same comment for the file upload component too.

image

dmbarry86:
I would expect the actions section to be showing the different callbacks as I'm selecting and removing files. it's just empty.

image


See checkbox component as an example that shows the actions a I interact with the component.
image

dmbarry86:
@Sowbhagya-ipg, just checking, is it intended that the component has a fixed width and height?

image

dmbarry86:
Is the dropzoneText prop actually working? I changed it here but it doesn't show up and still says the default 'drag and drop infographic image or click'. If we don't intend to change it then no need for prop. If we do intend to change it and expose prop then the storybook needs updating so the component updates as I change the prop. This applies to all props, not just this one. Idea of storybook is I can itneract with the props and see the changes.

image

dmbarry86 (changes requested):
See comments above.

Sowbhagya-ipg:
> Same comment regarding dark/light mode also applies for the file upload component which was previously merged.
> 
> Dark image
> 
> Light image

https://www.figma.com/proto/B75H0DFTEiUFjJtehIKa8M/Model-Manager?node-id=228%3A28915&scaling=contain&page-id=0%3A1&starting-point-node-id=228%3A28915

You can find the dark theme here as 10th task flow.  Implementing the same color and font to all dark theme drop zones will be ideal

Sowbhagya-ipg:
> @Sowbhagya-ipg, just checking, is it intended that the component has a fixed width and height?
> 
> image

Yes, The thumbnail image should preview the size exactly what the user is going to see on the cards page.

dmbarry86 (approved):

PR #506, opened 2023-01-26
Feature/wizard

mattcorner:
Release notes:
- New suite of wizard components that deliver a multi step form experience. Example usage: 
```
<Wizard title="Example wizard">
  <WizardSteps activeStep={1}>
    <WizardStep label="Step 1" helperText="Helper Text 1" />
    <WizardStep label="Step 2" helperText="Helper Text 2" />
    <WizardStep label="Step 3" helperText="Helper Text 3" />
  </WizardSteps>
  <WizardContent>
    // some content
  </WizardContent>
  <WizardActions>
    <CancelButton onClick={() => {}} />
    <BackButton onClick={() => {}} />
    <NextButton onClick={() => {}} />
  </WizardActions>
</Wizard>
```
- Improvements to the ThemeProvider including:
  - Addition of a new theme property `theme.layout.content.maxWidth`. This is to be used to set the maximum content width of relevant content e.g. Wizard.
  - Update to the `theme.palette.background.default` value. This is now a slight off-white/black colour to contrast with the `theme.palette.background.paper` value which is pure white/black.
  - Dark mode and light mode are now derived from the same base theme such that any custom layout styling applies to both.


dmbarry86:
Dark mode styling is not correct. See title, step labels, action bar etc.

image

dmbarry86:
Looks like your changes to the theme may have broken dark mode for all other components. For example, the select component below.

Main:
image

This branch:
image

src/Wizard/Wizard.js
```diff
@@ -0,0 +1,47 @@
...
+            mb: 3,
+            mt: 1,
+            mx: "auto",
+            width: "100%"
+          }}
+        >
+          {title}
+        </Typography>
+      ) : null}
+      {children}
```
dmbarry86, line 36: I think wizard should only be accepting known children and also rendering them in the correct order like MUI dialog behaves. I don't see the order documented anywhere either.

If I change the order to this then I can break the component layout.
image

image

src/Wizard/WizardActions/WizardActions.test.js
```diff
@@ -0,0 +1,18 @@
...
+    render(<WizardActions />);
+    expect(screen.getByRole("toolbar")).toBeInTheDocument();
+  });
+  it("is a row flexbox", () => {
+    render(<WizardActions />);
+    expect(screen.getByRole("toolbar")).toHaveStyle({
+      display: "flex",
+      flexDirection: "row"
+    });
+  });
```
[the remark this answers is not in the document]
dmbarry86: Presumably a similar story for the other components part of this PR that render children.

dmbarry86 (changes requested):
Few bits of feedback.

src/Wizard/Wizard.js
mattcorner: I did actually play around with Dialog when writing this and there are no restrictions in place that I can tell. You can render the Dialog children in any order you see fit, no warnings.

I can certainly throw a warning if the children appear out of order.

dmbarry86:
Something still not quite right with dark mode toggling. I have to toggle from light to dark, then back to light the to dark again for it to correctly show. For example, the copyright component here shows in black the first time I toggle to dark but then white (correct) the 2nd time.

https://user-images.githubusercontent.com/8976377/215504066-789c41f2-25d5-47df-bd66-914221b38956.mov

dmbarry86:
Same for wizard.


https://user-images.githubusercontent.com/8976377/215504590-63fd5aea-b082-492a-a10e-98a674228800.mov

dmbarry86 (changes requested):
Dark mode toggle button still not quite right. Requires two toggles for dark styling to be applied.

dmbarry86 (approved):

PR #507, opened 2023-01-26
Model button component

dmbarry86:
Contributes to TD-1273.

Added a new model button component as per https://www.figma.com/file/B75H0DFTEiUFjJtehIKa8M/Model-Manager?node-id=1451%3A53435&t=WpgDwwhWuRpQtzje-1

image

Component accepts props for disabled, icon, label, onClick and status. Styled as per Figma. Border color is set based on the status prop (none, error, warning or success). Hover effect changes border and icon to blue. Supports dark and light mode.

There will be a separate PR for the ModelButtonGroup component which will accept this ModelButton as children.

### DIFF
--- a/src/ModelButton/ModelButton.js
+++ b/src/ModelButton/ModelButton.js
@@ -1,0 +1,129 @@
+import * as React from "react";
+
+import { Box, IconButton, Typography } from "@mui/material";
+
+import PropTypes from "prop-types";
+import { useTheme } from "@mui/material/styles";
+
+/**
+ * Model button component to display a button with an icon, label and border color based on the model status
+ */
+export default function ModelButton({
+  disabled = false,
+  icon = null,
+  label = "",
+  onClick = () => {},
+  status = "none"
+}) {
+  // use theme hook
+  const theme = useTheme();
+
+  // set border color based on status
+  let borderColor = theme.palette.text.secondary;
+  if (status === "error") {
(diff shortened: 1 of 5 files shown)


mattcorner:
Is the expectation that when the button is disabled, the coloured outline remains? This feels a little weird to me. Its not immediately distinguishable as disabled.

dmbarry86:
> Is the expectation that when the button is disabled, the coloured outline remains? This feels a little weird to me. Its not immediately distinguishable as disabled.

I don't think there's a Figma for that case. Maybe @Sowbhagya-ipg can confirm whether disabled state should show grey or with the red/green/amber etc.

Sowbhagya-ipg:
@dmbarry86  @mattcorner  Check the Icon Main Components Page in Figma. You can find the states and colours used for icons. Also you can find it for dark theme.

mattcorner:
Some deviations from the Figma:

- [ ] Disabled state should also apply to the label and outline as well as the icon.
- [ ] Hover should apply a faded blue background
- [ ] Label font-size should be 13px.

mattcorner (changes requested):
Just some styling stuff to resolve. Code and functionality looks good.

mattcorner (changes requested):
Don't think the disabled font styling is being applied to label, other than that looks good.

mattcorner (approved):

PR #519, opened 2023-02-08
Feature/wizard max width

mattcorner:
Release notes:
* New `maxWidth` property on the `<Wizard />` component which allows overriding the default maximum content width of 1152px.

New api looks as follows, where maxWidth can be a string or a number. Behind the scenes it employs the same ThemeProvider override that we are using in place of this fix in some apps.

```
<Wizard maxWidth="1000px" />
<Wizard maxWidth={1000} />
```

### DIFF
--- a/src/Wizard/Wizard.js
+++ b/src/Wizard/Wizard.js
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import { Box, Typography } from "@mui/material";
+import { Box, ThemeProvider, Typography } from "@mui/material";
 
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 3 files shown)


src/Wizard/Wizard.js
```diff
@@ -45,8 +69,23 @@ Wizard.propTypes = {
    * Children. Should be WizardSteps, WizardContent, and WizardActions components in that order.
    */
   children: PropTypes.node,
+  /**
+   * Maximum width of the content. This includes the title, steps, and content, but not the actions.
```
dmbarry86, line 73:
```suggestion
   * Maximum width of the content. This includes the title, steps, and content, but not the actions. Default is 1152px.
```

dmbarry86:
Should probably show the default maxWidth value in the default story.

image

dmbarry86 (changes requested):
Looks good. Just a few small suggestions to show the default in the doc and add a test to confirm existing default behaviour.

dmbarry86 (approved):

PR #522, opened 2023-02-23
logo file to base64

m4manjesh:
- Added new SidebarLogo component to render base64 string instead of image file


jegasriskantha (approved):
LGTM

PR #523, opened 2023-02-23
Change Failed to Errored for in StatusLabel/StatusMessage

mattcorner:
Changes:
* Change Failed to Errored for in StatusLabel/StatusMessage to avoid mixing up with the "Failed" status of a metric.

Old:
image

New:
image

### DIFF
--- a/src/Status/statuses.js
+++ b/src/Status/statuses.js
@@ -40,7 +40,7 @@ const statuses = {
     },
     label: {
       color: red[700],
-      text: "Failed"
+      text: "Errored"
     }
   },
   "not-ready": {


dmbarry86:
Should line 36 also not be updated to say errored instead of failed?

image

dmbarry86 (commented):
Unsure about your intentions here. Are you still keeping everything in the database to still say 'Failed' but displaying as 'Errored' in the front-end? Might get confusing.

mattcorner:
> Should line 36 also not be updated to say errored instead of failed?
> 
> image

Can do, but I'd need to make further changes in our apps to cast the "failed" status we store, to "errored" status. I was treating this as a display only change.

dmbarry86 (changes requested):
Suggest this is held off until there's time to make all of the changes for consistency.

PR #524, opened 2023-02-24
side app bar for UI

m4manjesh:
Closes https://sce.myjetbrains.com/youtrack/issue/TD-1380/Create-new-react-ui-component-for-apps-menu

- UI of new sidebar app popper
- This will be integrated with this task https://sce.myjetbrains.com/youtrack/issue/TD-1379/Create-new-react-ui-component-for-header


src/SideAppbar/AppLogo.js
```diff
@@ -0,0 +1,34 @@
...
...MxNjYtNWIxYy00MTg2LWI5MDctNzlmMDVlZjRmNDE4IiBkPSJtNzMuOTcuMDJoLTEwLjY0Yy0xLjExLDAtMi4wMi44My0yLjAyLDEuODV2MTIuNjNjMCw1LjMxLS4zLDkuOTYtLjksMTMuOTgtLjA1LjMzLS4xLjY4LS4xNSwxLjAxLS43LDQuMjMtMS45NCw4LjA4LTMuNjksMTEuNTQtMS43NywzLjQ4LTQuMTYsNi43OC03LjE5LDkuOS0yLjU4LDIuNjctNC4zNyw1LjA3LTkuMDcsNy43MS0uNTUuMzEtMS4xNC41LTEuNzUuNTgtLjE5LjAzLS4zOC4wNC0uNTYuMDQtLjE5LDAtLjM4LS4wMS0uNTctLjA0LS42MS0uMDgtMS4yLS4yNy0xLjc1LS41OC00LjctMi42NC02LjQ5LTUuMDQtOS4wNy03LjcxLTMuMDMtMy4xMi01LjQyLTYuNDItNy4xOS05LjktMS43NS0zLjQ2LTIuOTktNy4zMS0zLjY5LTExLjU0LS4wNS0uMzMtLjEtLjY4LS4xNS0xLjAxLS42LTQuMDItLjktOC42Ny0uOS0xMy45OFYxLjg3QzE0LjY4Ljg1LDEzLjc3LjAyLDEyLjY2LjAySDIuMDJDLjkuMDIsMCwuODUsMCwxLjg3djguOThjMCw3LjI5LjM1LDEzLjczLDEuMDUsMTkuMzEuNyw1LjU5LDIuMTIsMTAuNzQsNC4yNCwxNS40NiwyLjExLDQuNzIsNS4wOSw5LjIsOC45MywxMy40MywyLjE4LDIuMzksNC44NCw1LjAyLDcuNzksNy43NCwzLjM1LDMuMDksNi45Miw1Ljk0LDEwLjYyLDguNTlsMy4wNiwyLjE5YzEuNDMsMS4xLDMuMTcsMS4xLDQuNjEsMGwzLjA2LTIuMTljMy43LTIuNjUsNy4yNy01LjUsMTAuNjItOC41OSwyLjk1LTIuNzIsNS42MS01LjM1LDcuNzktNy43NCwzLjg0LTQuMjMsNi44Mi04LjcxLDguOTMtMTMuNDMsMi4xMi00LjcyLDMuNTQtOS44Nyw0LjI0LTE1LjQ2LjctNS41OCwxLjA1LTEyLjAyLDEuMDUtMTkuMzFWMS44N2MwLTEuMDItLjktMS44NS0yLjAyLTEuODVaIi8+PC9nPjwvZz48L3N2Zz4=";
```
mattcorner, line 8: General comment about icons. We should really have a separate component per icon, and rather than embed data:image strings, we should export them from Figma as SVG, and expose them as a MUI SvgIcon. e.g. this would become something like:

https://codesandbox.io/s/new-thunder-bj1sfp?file=/demo.tsx

This applies to the Virto icon, but also all the individual app icons. If we expose these through react-ui, then they can be reused in each app, as well as the app-store where they are currently hardcoded. If we ever want to change an icon, we then only have to do it in one place.

src/SideAppbar/SideAppbar.js
```diff
@@ -0,0 +1,130 @@
...
+                  boxShadow: theme.shadows[1]
+                },
+                boxShadow: "none",
+                cursor: "pointer",
+                mb: 2,
+                width: 120
+              })}
+              onClick={() => handleClick(app.name)}
+            >
+              <Box sx={{ px: 2, py: 1.5, textAlign: "center" }}>
```
mattcorner, line 86: Something not quite right happening here when the text is slightly too long:

image

src/SideAppbar/SideAppbar.js
```diff
@@ -0,0 +1,130 @@
+import { Box, Link, Paper, Typography } from "@mui/material";
+
+import AppLogo from "./AppLogo";
+import PropTypes from "prop-types";
+import React from "react";
+
+// SideAppbar component for app which displays logo, list of items and app version
+function SideAppbar({ logoLinkUrl = null, showLogo = true, onButtonClick }) {
```
mattcorner, line 8: It took me ages to find this file because of the naming. I think maybe AppLauncher would be a better name?

src/SideAppbar/SideAppbar.js
```diff
@@ -0,0 +1,130 @@
...
+  // use theme
+
+  // VIRTO app list
+  const appList = [
```
mattcorner, line 12: Do we really want to hardcode the app list here? Thinking long term, not all customers might have access to all apps. We should pass in the array of apps during integration I think?

Thoughts @dmbarry86?

We can then also benefit from passing in the href so that we can use <Link /> properly rather than doing onButtonClick={(app) => navigate(app)}

mattcorner (changes requested):

src/SideAppbar/SideAppbar.js
m4manjesh: Renamed

src/SideAppbar/SideAppbar.js
m4manjesh: Passing the URL as a prop now.

mattcorner (approved):

mattcorner:
Looks good Manjesh, please create a draft release in Github and add some release notes before merging.

m4manjesh:
> Looks good Manjesh, please create a draft release in Github and add some release notes before merging.

Sure

PR #539, opened 2023-03-01
Fix dark mode primary color

dmbarry86:
As I was reviewing the new app layout component integration with our apps I noticed that dark mode primary color was wrong. You can see sidebar items were wrong color compared to Figma and so was anything in our apps that was using primary color, e.g. the plots in results app.

Before:
image

Now:
image


m4manjesh (approved):
Thanks for doing this. LGTM

PR #549, opened 2023-03-27
Bugfix/app layout scroll area

mattcorner:
Resolves issue that the content area of the AppLayout does not have an explitly calculated width set. This meant that scrollbars appeared over the sidebar, rather than only in the content area:

Before:
image

After
image


dmbarry86 (approved):
LGTM. Would be worth checking how each app is currently handling scrolling in the content area. Potentially some small changes needed in each app now we handle it here.

dmbarry86:
I'll leave you to merge and release.

PR #550, opened 2023-03-30
Switch to relative app links

dmbarry86:
Contributes to TD-1604.

Now that each app is served from a sub-path relative to a base URL, rather than sub-domains, we no longer need to manage links for each app. Instead we can just pass a base URL and then fix the relative links for each app inside the React-UI component. This simplifies usage in each app and reduces the number of environment variables down to just one for the base URL rather than the current implementation requiring a variable for each app.

AppLauncher is the component that contains the links for each app in the dropdown drawer. This now has a new url prop which replaces appUrls. The URL links for each app are then fixed inside this component relative to the base URL. AppLayout is the top level component we actually use in each app and this propagates the baseUrl through to AppLauncher.

These app links here are the ones related to these changes.
image

### DIFF
--- a/src/AppLayout/AppLayout.js
+++ b/src/AppLayout/AppLayout.js
@@ -151,7 +151,7 @@ function AppLayout({
 
 AppLayout.propTypes = {
   /**
-   * App version to display in header.
+   * App name to display in header.
    */
   appName: PropTypes.string.isRequired,
   /**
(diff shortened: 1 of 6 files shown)


src/AppLayout/AppLayout.js
```diff
@@ -154,23 +154,14 @@ AppLayout.propTypes = {
    * App version to display in header.
```
m4manjesh, line 154: I know this is not part of this PR. I can you have updated the wrong documentation in AppHeader.js . It would be nice if you can update the same here as well.

m4manjesh (commented):
LGTM. A small text update suggestion.

PR #562, opened 2023-04-03
ModelButtonImage component 

nikitharani:
Created a ModelButtonImage component  as a part of https://sce.myjetbrains.com/youtrack/issue/TD-1615/New-System-in-Model-Manager

- ModelButtonImage component is used to display the model icon in the ModelButton component.
- This component will be integrated into Apps where ModelButton Component is used

image

After Integrating, it looks like in the image.

Created this component to have a generic img style for all the buttons


src/ModelButtonImage/ModelButtonImage.stories.js
```diff
@@ -0,0 +1,17 @@
...
+  component: "ModelButtonImage",
+  title: "General/ModelButtonImage"
+};
+
+const DefaultTemplate = ({ src }) => {
+  return <ModelButtonImage src={src} />;
+};
+
+export const Default = DefaultTemplate.bind({});
+Default.args = {
```
m4manjesh, line 15: I would expect to see an option with ModelButton

src/ModelButtonImage/ModelButtonImage.js
```diff
@@ -0,0 +1,26 @@
...
+import React from "react";
+import { useTheme } from "@mui/material/styles";
+
+// ModelButtonImage component is used to display the model icon in the ModelButton component.
+export default function ModelButtonImage({ src }) {
+  const theme = useTheme();
+
+  return (
+    <>
+      {src && src.length > 0 ? (
```
m4manjesh, line 11: I don't think we need to add an extra length check here

src/ModelButtonImage/ModelButtonImage.js
```diff
@@ -0,0 +1,26 @@
...
+        <img
```
m4manjesh, line 12: use <Box component="img"

src/ModelButtonImage/ModelButtonImage.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+// wrapper function for ModelButtonImage
+function ModelImage({ src = ModelButtonSampleImg }) {
+  return <ModelButtonImage src={src} />;
+}
+
+// test suite for ModelButtonImage
+describe("ModelButtonImage", () => {
+  // component should show alt text
+  it("should show alt text  ", () => {
```
m4manjesh, line 15: The test name and test don't match, also i don't think we need to have a test for alt text. Do you mean render component?

src/ModelButtonImage/ModelButtonImage.js
m4manjesh, line 11: what will happen in ModelButton if no src? This component should always expect src so that make the prop required

src/ModelButtonImage/ModelButtonImage.js
```diff
@@ -0,0 +1,26 @@
...
+          style={{
+            filter: theme.palette.mode === "dark" ? "invert(1)" : "invert(0)"
+          }}
+        />
+      ) : null}
+    </>
+  );
+}
+
+ModelButtonImage.propTypes = {
```
m4manjesh, line 24: Add comments

m4manjesh, line 24: Add comments

m4manjesh (commented):
Few suggestions

m4manjesh (approved):
LGTM

PR #563, opened 2023-04-03
view toggle component

m4manjesh:


src/ViewToggleButton/ViewToggleButton.test.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+describe("Select", () => {
+  test("can select card view", () => {
+    render(<ViewToggleButtonWithState value="card" />);
+    const cardButton = screen.getByTestId("cardButton");
+    const tableButton = screen.getByTestId("tableButton");
+    expect(cardButton.getAttribute("aria-pressed")).toBe("true");
+    expect(tableButton.getAttribute("aria-pressed")).toBe("false");
+  });
+  test("can select table view", () => {
```
mattcorner, line 35: Can we have a test that checks if the user clicks on an option, that the onChange callback method is called?

mattcorner: See AlignVertical tests for reference.

src/ViewToggleButton/ViewToggleButton.js
```diff
@@ -0,0 +1,75 @@
...
+
+import PropTypes from "prop-types";
+
+/**
+ * Horizontal alignment button group component to toggle between card/table view
+ */
+export default function ViewToggleButton({
+  disabled = false,
+  onChange = () => {},
+  orientation = "horizontal",
```
mattcorner, line 14: I don't think we want to support orientation for this component.

src/ViewToggleButton/ViewToggleButton.test.js
m4manjesh: Test added

src/ViewToggleButton/ViewToggleButton.js
m4manjesh: removed

src/ViewToggleButton/ViewToggleButton.test.js
```diff
@@ -0,0 +1,50 @@
...
+    render(<ViewToggleButtonWithState value="table" />);
+    const cardButton = screen.getByTestId("cardButton");
+    const tableButton = screen.getByTestId("tableButton");
+    expect(cardButton.getAttribute("aria-pressed")).toBe("false");
+    expect(tableButton.getAttribute("aria-pressed")).toBe("true");
+  });
+  test("test onchange callback is called on click of an option", async () => {
+    const onChange = jest.fn();
+    render(<ViewToggleButtonWithState onChange={onChange} value="card" />);
```
mattcorner, line 45: I think what you've done here is written a test that proves that this ViewToggleButton component does not work as expected.

I should never be able to unselect all options, but what you're doing here is unselecting card, leaving nothing selected.

I've just tested this in storybook to confirm:

image

I suggest your ViewToggleButton only triggers the onChange callback when you have a value to set, and it is not equal to the current selection. See this section in the MUI docs for some detail: https://mui.com/material-ui/react-toggle-button/#enforce-value-set

mattcorner: I now expect a test that checks that the callback is not triggered when you click the currently selected button, as well as another test that checks that it is triggered when I click the other view button.

mattcorner: Also, you can now see the danger of just testing that a callback was called, rather than checking what args it was called with.

mattcorner (approved):

PR #571, opened 2023-04-12
Correct LabelChip font

mattcorner:
### Bug

LabelChip is showing the Roboto font rather than the Montserra font as raised by @Sowbhagya-ipg

### Resolution

We were wrapping each Chip in its own ThemeProvider to override its colour, but this also overrode everything in our React-UI ThemeProvider meaning we lost the font. I've removed the wrapping ThemeProvider and instead styled it using the sx prop so that we get the best of both worlds.

### Testing

* Open Storybook
* Inspect the element
* Check font shows as Montserra

https://user-images.githubusercontent.com/27866636/231391246-02f456ba-adf6-41f7-a899-ac0f4b3fae0d.mov

### DIFF
--- a/src/LabelSelector/LabelChip/LabelChip.js
+++ b/src/LabelSelector/LabelChip/LabelChip.js
@@ -1,4 +1,4 @@
-import { Chip, darken } from "@mui/material";
+import { Chip, alpha, chipClasses, darken } from "@mui/material";
 
 import PropTypes from "prop-types";
 import React from "react";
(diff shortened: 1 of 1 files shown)


nikitharani:
Before 
image

After 
image

The cancel icon color should match with text color

dmbarry86:
> Before image
> 
> After image
> 
> The cancel icon color should match with text color

Also looks like the Label chip height has increased, presumably as a result of the font change. You might want to check with @Sowbhagya-ipg what font size is expected in the label chips. Are we ok with this increased height or should font size be dropped?

mattcorner:
I've fixed the hover styling issue noted by @nikitharani.

I've measured the height of the chips and they remain consistent between main and this branch. (32px for normal, 24px for small). I think the screengrabs themselves were a little misleading in terms of size. The internal text itself is 1px taller, and a couple pixels wider however as a result of the font change. @Sowbhagya-ipg @dmbarry86

nikitharani (approved):
LGTM. I'll leave you to merge.

PR #580, opened 2023-04-25
Bugfix/light dark mode toggle

mattcorner:
## Bug

When setting an app to use dark mode, and then refreshing the page, the app reverted to light mode, and the localStorage value was overwritten to light.

## Resolution

We made some changes recently so that controlling the theme of the ThemeProvider using the `theme` property took priority over reading the localStorage value. This is required to support testing of light/dark mode, and is also how storybook syncs the dark/light mode with MUI. However, the default value for the controlled theme was "light" so it now always took priority. I've removed the default so that the theme is now decided using the following steps:
1. Is the theme being controlled? If so, use this value, otherwise...
2. Is the theme stored in localStorage? If so, use this value, otherwise...
3. Default to light.

## Changes

- Resolved bug as above.
- Migrated component and hook to typescript.
- Corrected the way we override component styles in the MUI Theme. Typescript highlighted that we were using an outdated API. We now use `component.componentName.styleOverrides`, rather than `overrides.componentName`
- Added unit tests to test all of the use cases for ThemeProvider.

## Not included
In Remix apps, we see a flicker from light mode to dark mode when refreshing in dark mode. This is because these are server-side rendered, but we cannot know the users preference until the page loads in the browser. I am writing a story to address in YouTrack.

## Testing notes

- Check that everything looks correct in storybook.
- Check that the tests pass.
- Install this branch in an application, and test that you can toggle to dark-mode, and then refresh the page. You should still be in dark-mode. Previously you would be reset to light mode. Confirm in the browser window, and also by looking at local storage.

https://user-images.githubusercontent.com/27866636/234552507-39ca0e31-f2bb-406a-88b7-6ddd51c2d00d.mov

## Release notes:
- Resolved bug with ThemeProvider where any page refresh resulted in the theme reverting to "light" mode.
- Migrated `ThemeProvider` and `useTheme` to typescript.

- [x] Assigned to me.
- [x] Relevant tags added.


nikitharani:
LGTM. You can merge when you are ready.

nikitharani (approved):

PR #581, opened 2023-04-26
Adds the ability to override the theme.palette.mode which automatically switches the logo text color

jegasriskantha:
relates to https://github.com/IPG-Automotive-UK/cloud-licensing-web/issues/274

- By default the light mode version of the IPG Logo has black text and the dark mode version has white text, for the cloud licensing web app the light mode version of the IPG Logo needs to have white text.  

## Changes
This PR enables the user to pass in an optional mode property which can be either "light" or "dark" which will allow the user to specifically set the version of the logo they required. 

I have also updated this component to be in Typescript. 

- A new mode property has been added to the IPG Logo component. When a mode has not been passed in to the component the theme.palette.mode is used (this preserves existing functionality) 

Maybe a short description of what is not included and will come in future work where appropriate.

## UI/UX
- Cloud Licensing Web Figma: https://www.figma.com/proto/O3IAobLTPcjIO1WaxaPod6/IPG-Cloud-Licensing?node-id=68-14265&starting-point-node-id=68%3A14265
- Specific logo requirement: 
image

- [x] Assigned to me.
- [x] Relevant tags added.


mattcorner:
This fix feels like a sticking plaster. 

As we've found out, this component has been written in a way that it is always intended to be placed on a light background in light mode, e.g. the Sidebar in VIRTO. In actual fact, how this component behaves in light and dark mode depends on where it is placed, so the logic for how to apply light/dark mode shouldn't live here, but in its parent. 

I'd propose that this component has a property to define the textColour. Then in Sidebar we pass in white or black as the text colour depending on the theme mode. This maintains current functionality.

Then in your use-case @jegasriskantha you can do the same (but opposite) logic in your CloudLicensing header.

Thoughts?

dmbarry86:
I am also not a fan of this implementation. I think there are better ways. It feels confusing to sometimes rely on the theme and sometimes to rely on the prop.

MUI already provides a way to wrap a component in a custom theme provider if needed. Can you not just use that and customise for your usecase?

The textColour prop suggestion above is also a good one. You could also just have two separate components; IPGLogoWhite, IPGLogoBlack etc.

jegasriskantha:
After a discussion with @dmbarry86 I think I will go ahead with the textColor prop as suggested by @mattcorner. I agree this is a much nicer and cleaner implementation. This does mean there will need to also be some updates to the components in react-ui that consume this component. 

@nikitharani I will close this PR and open a new one when I have made the suggested changes.

jegasriskantha:
For completeness @mattcorner and @dmbarry86  here is the improved implementation : https://github.com/IPG-Automotive-UK/react-ui/pull/583/files

PR #623, opened 2023-06-05
Bugfix/td 1565 dark mode model button overlay

mattcorner:
Contributes to [TD-1565](https://sce.myjetbrains.com/youtrack/issue/TD-1565/Dark-Mode-ModelButton-overlay)

## Changes

The previous implementation used a sudo element to sit over the bottom right corner of the model button, and replace the standard corner border with a cutout corner border. 

image

This worked well where the background was a consistent and known colour as shown here in light mode. However, in dark mode the background colour could differ, especially when the ModelButton sat in a Paper panel with a hover effect that modified the elevation. In dark mode that changes the background colour (https://mui.com/material-ui/react-paper/#elevation) unlike light mode. Hence the overlay became the wrong colour and you could see it. 

https://github.com/IPG-Automotive-UK/react-ui/assets/27866636/323ea2c4-26fd-418e-a028-a0fb8da8d936

The solution to this was to change our approach entirely. The new implementation uses two SVG representations of a button when it has children and when it doesn't. These SVG components are wrapped in a MUI SvgIcon component, and take in the `borderColor` and `backgroundColor` properties. This allows the parent button to control how it is displayed. This SVG is displayed within the Button, with the button icon displayed absolutely over the top. This solves the original issue, and now exactly matches the Figma design in terms of shape.

image

Then I noticed that in Figma, the icon displayed within the button should also change its hover color. This proved more difficult. Since the svg icons are hosted externally on the car-maker composer service, we display these SVG icons in an `` tag. This means that we cannot use the SVG fill property, or CSS fill property to colour it. 

Instead I've adopted a CSS filter approach using some of the learning described here: https://stackoverflow.com/a/43960991. Assuming that the original icon is black, it converts the target colour to a set of CSS filters in the format `filter: invert(a%) sepia(b%) saturate(c%) hue-rotate(θdeg) brightness(e%) contrast(f%);`. This can then be used to style the `` tag. 

Additionally I have introduced a cache to the `color2filter` utility function so that we need not recalculate the transform repeatedly. 

image

## UI/UX

I have created integration branches for [VIRTO.BUILD]() and [VIRTO.MODEL]() that you could spin up to test.

I have also given @Sowbhagya-ipg a brief demo, but for the sake of completeness, here are some screenshots.

image
image
image
image

image
image
image
image

image
image

## Testing notes

- I have updated the stories to use more representative icons. 
- I have added a story showing a ModelButton on top of a Paper card. The story allows you to modify the Paper elevation to prove that it works as expected in dark mode.
- I have added a test to ensure that a ModelButtonImage shows the correct colour when requested.


Sowbhagya-ipg:
Looks Good! @mattcorner

dmbarry86:
If I hover over the dropdown button, it just highlights the dropdown, which makes sense.

If I hover over the main button it highlights both the main button and the dropdown button. Should it not really only highlight the main button because they are two separate buttons with two separate actions? I see Figma is also implemented like you have here so I just want to confirm this with @Sowbhagya-ipg.

No hover.
image

Hover on dropdown button.
image

Hover on main button.
image

dmbarry86:
status prop in Storybook should be a dropdown not text field.

image

dmbarry86:
Missing prop descriptions in Storybook.

image

dmbarry86:
I notice that the ModelButton doesn't appear in the coverage report on the GitHub action. Potentially a few components are missing actually. Is this a result of the migration to Typescript?

image

dmbarry86 (changes requested):
New component code looks good. A few items of feedback though.

- [x] Confirm hover behaviour is intended
- [ ] Status prop in Storybook should be a dropdown
- [ ] Missing sections of Storybook doc
- [x] Typescript components missing from coverage report

Sowbhagya-ipg:
> If I hover over the dropdown button, it just highlights the dropdown, which makes sense.
> 
> If I hover over the main button it highlights both the main button and the dropdown button. Should it not really only highlight the main button because they are two separate buttons with two separate actions? I see Figma is also implemented like you have here so I just want to confirm this with @Sowbhagya-ipg.
> 
> No hover. image
> 
> Hover on dropdown button. image
> 
> Hover on main button. image

You may be right @dmbarry86. Now, I checked with other sites for this type of button behavior, and some have used this way and most others have hover interaction separately for each button. @mattcorner Can you keep it the same way as it is devel now, please?

PR #638, opened 2023-06-28
Fix for SVGIcon in latest @mui/material

mattcorner:
## Changes

- Upgrades to `@mui/material@5.3.16`
- Resolves issue with VirtoLogo was not honoured. This was because until now MUI did not support `<svg />` being a child of `<SvgIcon />`. As a result what we actually ended up with was an `svg` inside of an `svg`. Now that MUI is supporting this it broke the styling. There is no benefit for us to have `<SvgIcon>` and `svg` in our use cases so I have removed the nesting.

## UI/UX

Before with newest @mui/material:
image

After fix:
image

## Testing notes

Check `VirtoAppLayout`, `VirtoAppHeader`, `AppLauncher`. There should be no difference to main.


dmbarry86 (approved):
Updated icons look fine to me. Small question over dark mode inconsistencies but not related to this directly.

PR #659, opened 2023-07-13
Migrate Wizard components to typescript

mattcorner:
## Changes

* Migrated `Wizard`, `WizardContent`, `WizardSteps`, `WizardStep`, `WizardActions`, `NextButton`, `BackButton`, `CancelButton` to typescript.
* Turned off some unnecessary story controls. 

## UI/UX

No changes

## Testing notes

- Ensure component acts and behaves as before.
- Ensure docs are correctly parsed from the types
- Ensure types look correct


src/Wizard/WizardActions/CancelButton/CancelButton.types.ts
```diff
@@ -0,0 +1,10 @@
+export interface CancelButtonProps {
```
dmbarry86, line 1: Looks like the type definition for each of the 3 buttons is identical and could benefit from a shared type, presumably in the level above in wizard types.

dmbarry86: I guess 'ActionButtonProps'.

src/Wizard/WizardSteps/WizardStep/WizardStep.tsx
```diff
@@ -34,42 +36,16 @@ export default function WizardStep({
...
  */
-function WizardStepLabel({ label, helperText, errorText }) {
+function WizardStepLabel({
+  label,
+  helperText,
+  errorText
+}: Pick<WizardStepProps, "label" | "helperText" | "errorText">) {
   // get context state
-  const { active, completed } = useStepContext();
+  const { active, completed } = useStepContext() as StepContextType;
```
dmbarry86, line 48: Why do we need to use as here? Isn't this MUI hook already typed?

dmbarry86 (changes requested):
Functionality and doc looks good. Just a couple of small points about the types.

src/Wizard/WizardSteps/WizardStep/WizardStep.tsx
mattcorner: It is typed, but it is typed as `StepContextType | {}`. The documented reason for this in code is that technically if you use the `useStepContext` component not within its context provider, then you'll get an empty object. Makes sense. 

The problem is I have found no way to type-guard against an empty object, so I can't get typescript to acknowledge that I can access the `active` and `completed` props. MUI gets around this by the fact that their code is written in js, and the type are written alongside. 

Given that WizardStepLabel is always rendered within a `Step`, and a `Step` always [generates the context](https://github.com/mui/material-ui/blob/1b9c8ab5a7c0125fa411f46ffafba76f04fc6eae/packages/mui-material/src/Step/Step.js#L111-L120) I gave up sinking any more time into this.

[Typescript playground](https://www.typescriptlang.org/play?strictNullChecks=false&noImplicitReturns=false#code/C4TwDgpgBAysFgMIHsB28AewoF4oG8AoKEqAQwGNgBLANwgC4oAjZZAGwjNQG5jSKyALZhO8ACZNWHLr0IBfQoUGoAztmRgaaAIK4ClGvSbAATgFcIAGiiCRYiJKgAzMu1UR55VQS8AfWHgkNExgPkIAegioc1VqVABzKDR2EChgAAtodniIH2YIdmQAdxtVYQgMkqghMgBraFBIVQpTai0Y1HEIU3VucXSMsmwAeS1qXShqH2RndPBoOAQUdAgsG36oVGQNOabofEVCIA) if you care to have a play.

dmbarry86: https://www.typescriptlang.org/play?strictNullChecks=false&noImplicitReturns=false#code/C4TwDgpgBAysFgMIHsB28AewoF4oG8AoKEqAQwGNgBLANwgC4oAjZZAGwjNQG5jSKyALZhO8ACZNWHLr0IBfQoUGoAztmRgaaAIK4ClGvSbAATgFcIAGiiCRYiJKgAzMu1UR55VQS8AfWHgkNExgPkIAegioc1VqVABzKDR2EChgAAtodniIH2YIdmQAdxtVYQgMkqghMgBraFBIVQpTai0Y1HEIU3VucXSMsmwAeS1qXShqH2RndPBoOAQUdAgsG36oVGQNOabofEVqOYAKAHI7UQgJM6nUZPHdAEoCfhIorer9lraOuu3ijNHqg9EJYtgCoFliE1sA3rY0OoEfZro4AJLAACyw2geE02hBADpLg5xHx5EA

dmbarry86 (approved):

PR #662, opened 2023-07-19
Snackbar flex-wrap

dmbarry86:
Fixes an issue spotted in a cloud licensing pull request https://github.com/IPG-Automotive-UK/cloud-licensing-web/pull/298#issuecomment-1641880722.

## Changes

If the snackbar message was long or the screen width was small then the text was wrapping onto a newline, however so was the action button to close the dialog. The action button is now always on the right and vertically centre aligned with the message text.

## UI/UX

Before:
image

After:
image

## Testing notes

Open up storybook for Snackbar, set a long message or resize the screen width so the message wraps onto multiple lines and confirm close action button is as expected.

### DIFF
--- a/src/Snackbar/Snackbar.types.ts
+++ b/src/Snackbar/Snackbar.types.ts
@@ -6,12 +6,6 @@ import {
 export interface SnackbarProps {
   /**
    * Callback fired when the user clicks the action button.
-   *
-   * **Signature**
-   * ```
-   * function(event: object) => void
-   * ```
-   * _event_: The event source of the callback.
    */
   actionCallback?: (event: React.SyntheticEvent<HTMLElement>) => void;
   /**
@@ -28,17 +22,8 @@ export interface SnackbarProps {
   message: SnackbarContentProps["message"];
   /**
    * Callback fired when the component requests to be closed. Typically onClose is used to set state in the parent component, which is used to control the Snackbar open prop.
-   *
-   * **Signature**
-   * ```
-   * function(event, reason) => void
-   * ```
-   *
-   * _event_: The event source of the callback.
-   *
-   * _reason_: Can be: "timeout" (autoHideDuration expired), "clickaway", "action".
    */
-  onClose: MuiSnackbarProps["onClose"];
+  onClose?: MuiSnackbarProps["onClose"];
   /**
    * If true, Snackbar is open.
    */
--- a/src/Snackbar/Snackbar.types.ts
+++ b/src/Snackbar/Snackbar.types.ts
@@ -1,0 +1,50 @@
+import {
+  SnackbarProps as MuiSnackbarProps,
+  SnackbarContentProps
+} from "@mui/material";
+
+export interface SnackbarProps {
+  /**
+   * Callback fired when the user clicks the action button.
+   *
+   * **Signature**
+   * ```
+   * function(event: object) => void
+   * ```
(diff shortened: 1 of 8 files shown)


Sowbhagya-ipg:
> 

This is Good!

nikitharani (commented):
The changes LGTM. But the component is not migrated to TypeScript.

dmbarry86:
> But the component is not migrated to TypeScript.

Done. My 1 line bug-fix has now turned into 126 lines. Please ensure you review throughly again now as this not only affects the bug-fix.

src/Snackbar/Snackbar.types.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  SnackbarProps as MuiSnackbarProps,
+  SnackbarContentProps
+} from "@mui/material";
+
+export interface SnackbarProps {
+  /**
+   * Callback fired when the user clicks the action button.
+   *
+   * **Signature**
```
nikitharani, line 10: I think we can remove signatures in the typescript because the storybook infers them for us.

src/Snackbar/Snackbar.types.ts
```diff
@@ -0,0 +1,50 @@
...
+   */
+  autoHideDuration?: MuiSnackbarProps["autoHideDuration"];
+  /**
+   * The message to display.
+   */
+  message: SnackbarContentProps["message"];
+  /**
+   * Callback fired when the component requests to be closed. Typically onClose is used to set state in the parent component, which is used to control the Snackbar open prop.
+   *
+   * **Signature**
```
nikitharani, line 32: same here we can remove signatures in the typescript because the storybook infers them for us.e

src/Snackbar/Snackbar.types.ts
```diff
@@ -0,0 +1,50 @@
...
+   * ```
+   * function(event, reason) => void
+   * ```
+   *
+   * _event_: The event source of the callback.
+   *
+   * _reason_: Can be: "timeout" (autoHideDuration expired), "clickaway", "action".
+   */
+  onClose: MuiSnackbarProps["onClose"];
```
nikitharani, line 41: Previously, onClose was an optional prop, but now it's a required prop was it an intentional change? I am asking this because we have autoHideDuration in such cases we don't want onClose to be a required prop right? please correct me if I am wrong,

nikitharani (commented):
LGTM. Just a small change is required.

src/Snackbar/Snackbar.types.ts
dmbarry86: This wasn't intentional no so good spot. MUI itself has this optional I think so we should probably follow suit. However, I think your understanding of autoHideDuration is not quite right. When autoHideDuration is set it just automatically calls onClose function after the given number of milliseconds. It doesn't actually close the dialog, that is the job of the onClose and the parent component which is controlling the open state. See the help comments for autoHideDuration prop: "The number of milliseconds to wait before automatically calling the onClose function. onClose should then set the state of the open prop to hide the Snackbar. This behavior is disabled by default with the null value." I don't actually see it possible to use the component in a controlled manner without using open and onClose props together but we can keep onClose optional to follow MUI and ensure nothing breaks.

nikitharani (approved):
I'll leave you to merge and release.

PR #695, opened 2023-08-29
TD 2069 vehicle select dialog

nikitharani:
Closes [TD-2069](https://sce.myjetbrains.com/youtrack/issue/TD-2069/Create-add-vehicle-dialog)
Figma: https://www.figma.com/proto/B75H0DFTEiUFjJtehIKa8M/VIRTO.MODEL?node-id=522-57017&scaling=scale-down&page-id=0%3A1&starting-point-node-id=228%3A28915

## Changes
- Created a new VehicleSelectDialog component
- Enables the "Add Vehicle" upon valid vehicle selection
- Added callback to return the currently selected vehicles
- Upgraded to node version 18 and upgraded the actions/setup-node action to the latest version
- Moved ActionDialog, DialogTitle, and VehicleSelectDialog components under the Dialog category from General
- Renamed selected vehicles to value, all gates to gates, on vehicle change to onChange, allVehicles to variants to be consistent with vehicleSelect Component
- Added a new prop disableCloseOnSelect to the AutoComplete Component
- Enabled disableCloseOnSelect on multi-select Autocompletes inside VehicleSelect Component
- Because of the disableCloseOnSelect prop which prevents the dropdown from closing when an option is selected (VehicleSlect Component test failed). Modified tests to explicitly close the dropdown after selecting an option using userEvent.click.

## UI/UX
image

Enables the "Add Vehicle" upon valid vehicle selection
image

 The callback returns the currently selected vehicles
image

### DIFF
--- a/src/VehicleSelectDialog/VehicleSelectDialog.stories.tsx
+++ b/src/VehicleSelectDialog/VehicleSelectDialog.stories.tsx
@@ -1,13 +1,10 @@
 import { Meta, StoryFn } from "@storybook/react";
 import React, { MouseEventHandler } from "react";
-import {
-  SelectedVehicle,
-  VehicleSelectProps
-} from "../VehicleSelect/VehicleSelect.types";
 
 import { Button } from "@mui/material";
+import { CombinedVehicleProps } from "./VehicleSelectDialog.types";
+import { SelectedVehicle } from "../VehicleSelect/VehicleSelect.types";
 import VehicleSelectDialog from "./VehicleSelectDialog";
-import { VehicleSelectDialogProps } from "./VehicleSelectDialog.types";
 import { action } from "@storybook/addon-actions";
 import { useArgs } from "@storybook/client-api";
 
@@ -21,7 +18,7 @@ const meta: Meta<typeof VehicleSelectDialog> = {
 
 export default meta;
 
-const Template: StoryFn<VehicleSelectDialogProps> = args => {
+const Template: StoryFn<CombinedVehicleProps> = args => {
   const [open, setOpen] = React.useState(false);
   const handleClickOpen = () => setOpen(true);
   const handleCancel: MouseEventHandler<HTMLButtonElement> = args => {
@@ -34,7 +31,7 @@ const Template: StoryFn<VehicleSelectDialogProps> = args => {
   };
 
   // useArgs is a hook that returns the current state of the args object
-  const [{ selectedVehicles }, updateArgs] = useArgs<VehicleSelectProps>();
+  const [{ selectedVehicles }, updateArgs] = useArgs<CombinedVehicleProps>();
 
   // update the args object with the new selectedVehicles value
   React.useEffect(() => {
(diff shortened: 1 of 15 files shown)


evoinea:
@Sowbhagya-ipg can you please have a look on UI/UX changes?

Sowbhagya-ipg:
> @Sowbhagya-ipg can you please have a look on UI/UX changes?

thanks for tagging me. I will do it now

src/VehicleSelectDialog/VehicleSelectDialog.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const saveButton = screen.getByText("Save");
+  fireEvent.click(saveButton);
+
+  // check that the onSaveClick function was called
+  expect(mockOnSaveClick).toHaveBeenCalledWith(selectedVehicles);
+});
+
+test("disables the Save button when any field is empty", () => {
+  // modify the selected vehicles to have an empty gate field
+  const modifiedVehicles = [...selectedVehicles];
```
evoinea, line 85:
```suggestion
  const modifiedVehicles = [...selectedVehicles];
  
```

src/VehicleSelectDialog/VehicleSelectDialog.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+  // empty the gate field
+  modifiedVehicles[0].gate = "";
+
```
evoinea, line 88: Please add missing comment

src/VehicleSelectDialog/VehicleSelectDialog.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+const meta: Meta<typeof VehicleSelectDialog> = {
+  component: VehicleSelectDialog,
+  title: "General/VehicleSelectDialog"
+};
+
+export default meta;
+
+const Template: StoryFn<CombinedVehicleProps> = args => {
+  const [open, setOpen] = React.useState(false);
+
```
evoinea, line 23: Missing comments

src/VehicleSelectDialog/VehicleSelectDialog.stories.tsx
```diff
@@ -0,0 +1,115 @@
...
+  React.useEffect(() => {
+    updateArgs({ selectedVehicles });
+  }, [selectedVehicles, updateArgs]);
+
+  // callback for when the selected vehicles change
+  const onVehicleChange = (selectedVehicle: SelectedVehicle[]) => {
+    updateArgs({ selectedVehicles: selectedVehicle });
+    action("onVehicleChange")(selectedVehicle);
+  };
+
```
evoinea, line 47: Missing block comment

evoinea (changes requested):
Few small changes required.

evoinea:
I've deleted the Model Year, but this didn't clear the vehicle variant and the gate; it just disabled them:
image

dmbarry86:
I think the interface for this component has been designed wrong and makes it harded to use than it should be. The dialog itself should be handling the local state when user selects items. You don't need to expose onChange callback from the dialog and manage state outside of the component and then pass in value. You only need onCancelClick and onSaveClick callbacks in this component. Component itself listens to onChange in the VehicleSelect and keeps state in synch then when onSaveClick is called you just need to pass out what the currently selected value is. Otherwise everywhere that uses this component then needs to handle onChange and have external value state. I don't think we ever need to open the dialog with initial values selected so I don't think we need values and onChange props. It's a much simpler interface without them.

dmbarry86:
these props are still showing for me in Storybook

image

dmbarry86:
the props are also still included in the type definition

image

dmbarry86 (changes requested):
- [ ] remove onChange and value from storybook
- [ ] remove onChange and value from types
- [ ] ensure state is reset when dialog is opened

dmbarry86 (approved):

PR #697, opened 2023-08-31
Fix spread syntax bundling error with VehicleSelect

dmbarry86:
Contributes to [TD-1435](https://sce.myjetbrains.com/youtrack/issue/TD-1435/Multiselection-of-Vehicle-Variants-in-Data-Model-upload)

## Changes

When trying to make use of the new VehicleSelect component in one of our apps I noticed we were getting this weird behaviour.
image

I've tracked it down to an issue with the bundler and known issue https://github.com/developit/microbundle/issues/945.

We were making use of spread syntax to get unique values in an array
```
const selectedProjects = [
    ...new Set(value.map(vehicle => vehicle.projectCode))
  ];
```

When bundled this was building down to this which is not the same.
```
var selectedProjects = [].concat(new Set(value.map(function (vehicle) {
    return vehicle.projectCode;
  })));
```

You can check the bundled code using `npm run build` and then inspecting `dist/index.js` and finding the VehicleSelect function.

This is what it used to be
image

I've switched away from using the spread syntax. In the future we may switch to a different bundler but for now this workaround is fine. We should just keep in mind for any future enhancements.

## Testing notes

The automated tests and type checks cover the change but you could also manually test the VehicleSelect component in storybook `npm run storybook`

- [ ] Branch has been run in docker. N/A
- [ ] Appropriate tests have been added. N/A already covered by existing tests.

### DIFF
--- a/src/VehicleSelect/VehicleSelect.tsx
+++ b/src/VehicleSelect/VehicleSelect.tsx
@@ -3,6 +3,11 @@ import { Box } from "@mui/material";
 import React from "react";
 import { VehicleSelectProps } from "./VehicleSelect.types";
 
+// function to return unique sorted array of values from a string array
+function uniqueSortedArray(values: string[]) {
+  return values.filter((item, i, ar) => ar.indexOf(item) === i).sort();
+}
+
 // component to select a vehicle
 function VehicleSelect({
   flexDirection = "column",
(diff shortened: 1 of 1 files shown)


nikitharani (approved):
LGTM

PR #698, opened 2023-09-01
Fixed issue disabling vehicle select with empty value

dmbarry86:
Contributes to [TD-2065](https://sce.myjetbrains.com/youtrack/issue/TD-2065/Create-AddVehicle-component)

## Changes

When making use of the recently released VehicleSelect component in DATA https://github.com/IPG-Automotive-UK/virto/pull/111 it was spotted that on first render in DATA the model year and variant fields were not disabled. Gate was correctly disabled but those two were not. The logic should be that all subsequent fields are disabled until the prior field is disabled, e.g. model year is disabled until a valid project has been selected etc. The issue was the fact that the component was dealing with empty value.

This was not being dealt with:
```
value = [];
```

But this was being dealt with:
```
value = {
        _id: "",
        gate: "",
        modelYear: "",
        projectCode: "",
        variant: ""
      }
```

The default for value is actually an empty array `value: []`. We didn't spot this because the default prop value in the story was an empty object not and empty array. We also didn't have any tests to check the logic for disabling the fields. I've fixed the issue and have improved the test coverage to check for all posibilities.

## Testing notes

The new tests I have added should provide full coverage for this case and all other variants of value. You could also manually check by opening storybook and checking the default story for VehicleSelect now correctly only shows project code field enabled and all others disabled. I updated the default story to correctly show with value as empty array.
`npm run storybook`

### DIFF
--- a/src/VehicleSelect/VehicleSelect.stories.tsx
+++ b/src/VehicleSelect/VehicleSelect.stories.tsx
@@ -38,15 +38,7 @@ export const Default = {
     flexDirection: "column",
     flexWrap: "nowrap",
     gates: ["Gate 1", "Gate 2", "Gate 3"],
-    value: [
-      {
-        _id: "",
-        gate: "",
-        modelYear: "",
-        projectCode: "",
-        variant: ""
-      }
-    ],
+    value: [],
     variants: [
       {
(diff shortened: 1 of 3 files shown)


nikitharani (approved):
Tested LGTM.

PR #748, opened 2023-10-03
Playwright runner fails on GitHub actions

dmbarry86:
## Changes

Fixes the issue with running Playwright tests in GitHub action runner sometimes failing. The way we were running the test was flawed and we just got lucky that they passed locally because local machines are higher performance than GitHub action runners. We were running tests against dev mode storybook and we were trying to wait for storybook to be ready before running the tests but storybook dev mode will return a http 200 response before it has actually completed bulding. You see this with `npm run storybook` where it opens the browser quite quickly but then continues to build before the app is actually responsive. The fix I've implemented is to switch to using a compiled static build of storybook and properly waiting for the build to be served before running the Playwright tests. Not only does this fix the bug with running tests but it also more closely resembles how the components are built and served in production.

Note that in running the tests a few times locally I've uncovered that the (non-playwright) tests for VehicleSelect component are flaky. Sometimes they run, sometimes they timeout. I've raised with the respective author of those tests and they will be fixed separately to this PR.

In doing this I also noticed that we were not running type checks in the actions. I've taken the opportunity to run each check as a separate line item with user friendly names so we can inspect the action logs a bit better.

## Dependencies

N/A

## UI/UX

N/A

## Testing notes

Check that the following scripts run locally:

`npm run test:playwright`
`npm run test`

You can inspect the latest GitHub action runner logs [here](https://github.com/IPG-Automotive-UK/react-ui/actions/runs/6396496864/job/17362381917?pr=748).

- [ ] Branch has been run in docker. N/A


VictorCLLucas:
I ran `npm run test` and it seems to be getting stuck when it gets to the playwright tests. I waited for about 8 minutes before taking the screen shot but it just stays there. 
image

VictorCLLucas:
When running `npm run test:playwright` it gets stuck here
image

dmbarry86:
It was a Windows issue you were seeing Victor. I've switched the way we serve storybook over to something that should now be OS agnostic.

dmbarry86:
I've also now updated the action so that it runs on windows, linux and mac.

VictorCLLucas:
These are now passing locally for me.
Just one point it seems to be a bit overkill to run this in 3 separate os's is this really the best way? Not to mention it takes 3 times as long.

VictorCLLucas (approved):

dmbarry86:
> These are now passing locally for me.
> Just one point it seems to be a bit overkill to run this in 3 separate os's is this really the best way? Not to mention it takes 3 times as long.

I would suggest we keep the 3 OS for now as we've clearly seen here and with the other plot PR that OS can have an impact on the way Playwright runs. We have a mix of Windows and Mac for developers and deployment is via Linux so I think we should keep the 3 OS for now.

PR #751, opened 2023-10-05
Bug/vehicle select component test fix

nikitharani:
BugFix for vehicle select component tests
 
## Changes
- Using waitFor in the test to wait for the condition to be met without causing timeout errors.
- Added Jest timeout that increases the timeout for the test suite.

## Testing notes
On the "npm run test" all tests should pass without any errors


dmbarry86:
I'm not 100% convinced about this "fix". We shouldn't need to be increasing timeouts to 10s for a callback. Thoughts on this @mattcorner ?

mattcorner:
> I'm not 100% convinced about this "fix". We shouldn't need to be increasing timeouts to 10s for a callback. Thoughts on this @mattcorner ?

I've spent a little bit of time having a look. On my Mac the VehicleSelect tests take 5-7 seconds, which is the longest of any of the test suites in react-ui. 

The onChange seems to be the longest test at 1-2 seconds. This test does quite a lot as it works it way through the entire vehicle selector clicking all the menus. 

I can't really see another solution at the moment other than accepting that this test takes longer on less performant hardware.

PR #762, opened 2023-10-26
Enhancement - Options to allow key value array

syedsalehinipg:
## Changes

- Implemented functionality to support options arrays consisting of key-value pairs

## Dependencies

N/A

## Testing notes

Example:

`[
      { key: "option-a", value: "Option A" },
      { key: "option-b", value: "Option B" }
    ]`

### DIFF
--- a/src/Autocomplete/Autocomplete.stories.tsx
+++ b/src/Autocomplete/Autocomplete.stories.tsx
@@ -1,7 +1,7 @@
+import { AutocompleteProps, KeyValueOption } from "./Autocomplete.types";
 import { Meta, StoryFn, StoryObj } from "@storybook/react";
 
 import Autocomplete from "./Autocomplete";
-import { AutocompleteProps } from "./Autocomplete.types";
 import React from "react";
 import { action } from "@storybook/addon-actions";
 import { useArgs } from "@storybook/client-api";
@@ -23,7 +23,7 @@ function isReadOnlyArray<T>(
 }
 
 const Template: StoryFn<
-  AutocompleteProps<string, boolean | undefined>
+  AutocompleteProps<string | KeyValueOption, boolean | undefined>
 > = args => {
   const [{ value, multiple }, updateArgs] =
     useArgs<AutocompleteProps<string, boolean | undefined>>();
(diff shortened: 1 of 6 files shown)


dmbarry86:
@syedsalehinipg maybe Nikitha didn't mention during the handover, but the strategy we have RUI TypeScript migration is that all new components should be written in TypeScript and existing components get migrated when they are touched so that eventually we move everything over bit by bit.

Also, where are you planning to use this Select component? In general we should be moving to the Autocomplete.

dmbarry86:
I would also expect some updated tests to cater for this new syntax as well as ensuring the old syntax hasn't been broken.

dmbarry86 (changes requested):
- Migrate to TypeScript
- Missing tests for new behaviour
- Question over Select usage vs Autocomplete

syedsalehinipg:
> @syedsalehinipg maybe Nikitha didn't mention during the handover, but the strategy we have with RUI TypeScript migration is that all new components should be written in TypeScript and existing components get migrated when they are touched so that eventually we move everything over bit by bit.
> 
> Also, where are you planning to use this Select component? In general we should be moving to the Autocomplete.

This was going to be for the Measures dropdown in the Unit Conversion. And regarding changing it to typescript I did think that was the idea but it's just the timeframe I planned to complete this in could be delayed if I get side tracked with changing components to typescript if I have to make any small change, hence why I didn't. But if you are okay if I don't meet what's planned then I am okay doing that.

syedsalehinipg:
> @syedsalehinipg maybe Nikitha didn't mention during the handover, but the strategy we have with RUI TypeScript migration is that all new components should be written in TypeScript and existing components get migrated when they are touched so that eventually we move everything over bit by bit.
> 
> Also, where are you planning to use this Select component? In general we should be moving to the Autocomplete.

Should I be using the autocomplete instead? For the unit conversions I think its easier to have dropdown for a user point of view, no? 

If so I can scrap this.

dmbarry86:
> I think its easier to have dropdown for a user point of view, no?

The autocomplete is a dropdown but is also allows to search, thus a much improved UX over Select. Run Storybook and give the Autocomplete component a try.

syedsalehinipg:
> > I think its easier to have dropdown for a user point of view, no?
> 
> The autocomplete is a dropdown but is also allows to search, thus a much improved UX over Select. Run Storybook and give the Autocomplete component a try.

Oh yes, I see it. It is better for sure. I might have to make the same changes here too if it doesn't support key value pair options.

dmbarry86 (approved):

PR #766, opened 2023-11-08
Deploy js alongside ts

dmbarry86:
## Changes

We want to make some component enhancements in RUI and consume them in our old JS based apps. Since RUI v5 we've only been shipping source code and a lot of components are now TS. The updates I've made here now mean that we compile the TS and ship JS alongside. This means we can now use all components from RUI 5 in JS and TS apps. I've also updated to compile JSX to JS as we have issues consuming JSX in our older create-react-app JS apps. In doing this I came across an issue with our resize-observer polyfill not playing nice in TS so I've replaced it with a maintained and recommend equivalent that does compile well.

Large line diff is mostly due to gitignore and eslintignore additions to ignore the build files.

## Dependencies

N/A

## UI/UX

N/A

## Testing notes

I've created a pre-release build and have integrated it in to VIRTO.DATA, VIRTO.VEHICLE, VIRTO.ID and VIRTO.TEST for testing. All is well. You can also test VIRTO.DATA/VEHICLE/ID in dev and prod mode on that branch https://github.com/IPG-Automotive-UK/virto/tree/dependencies/rui-5-in-js-apps. I added VIRTO.TEST so I could check I didn't break anything for a TypeScript Remix app.


package.json
```diff
@@ -1,16 +1,18 @@
...
   "repository": "IPG-Automotive-UK/react-ui",
-  "main": "src/index.ts",
+  "main": "src/index",
   "engines": {
     "node": ">=10"
   },
   "scripts": {
+    "build": "tsc --build --clean && tsc",
     "postinstall": "npx playwright install --with-deps",
+    "prepublishOnly": "npm run build",
```
dmbarry86, line 15: This ensures typescript compilation gets run just before we publish to npm.

package.json
```diff
@@ -1,16 +1,18 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "5.5.0",
+  "version": "5.5.1-4",
   "description": "React UI component library for IPG web applications",
   "author": "IPG-Automotive-UK",
   "license": "MIT",
   "repository": "IPG-Automotive-UK/react-ui",
-  "main": "src/index.ts",
+  "main": "src/index",
```
dmbarry86, line 8: Removed file extension so consuming apps can be js or ts import.

tsconfig.json
```diff
@@ -1,17 +1,22 @@
 {
   "compilerOptions": {
-    "allowJs": true,
-    "baseUrl": ".",
+    "allowJs": false,
```
dmbarry86, line 3: Not sure why this was ever set to be honest. I suspect copy/paste of tsconfig from somewhere without really knowing what is happening. We don't need to be touching any js.

tsconfig.json
```diff
@@ -1,17 +1,22 @@
...
     "lib": ["DOM", "DOM.Iterable", "ES2019"],
     "moduleResolution": "node",
     "module": "ES2022",
-    "noEmit": true,
     "resolveJsonModule": true,
     "strict": true,
     "target": "ES2019"
-  }
+  },
+  "include": ["src/**/*.tsx", "src/**/*.ts", "src/**/*.jsx"],
```
dmbarry86, line 15: As well as compiling TypeScript we also need to compile JSX to JS as our old apps can't deal with JSX imports.

package.json
```diff
@@ -30,6 +32,7 @@
     "@babel/preset-react": "^7.22.15",
     "@emotion/react": "^11.11.1",
     "@emotion/styled": "^11.11.0",
+    "@juggle/resize-observer": "^3.4.0",
```
dmbarry86, line 35: had to switch our polyfill over from "resize-observer-polyfill" to this as the old one didn't play nicely with ts compilation. Seems to no longer be maintained and this library was recommended by others who came across the same issue. Slots in like for like.

.eslintignore
```diff
@@ -2,4 +2,434 @@ build/
 dist/
 node_modules/
 .snapshots/
-*.min.js
\ No newline at end of file
+*.min.js
+
+# js build files where a ts source file exists
```
dmbarry86, line 7: Ensure we ignore all the temp build files from eslint as they don't conform.

.gitignore
```diff
@@ -29,3 +29,433 @@ coverage
 /test-results/
 /playwright-report/
 /playwright/.cache/
+
+# js build files where a ts source file exists
```
dmbarry86, line 33: Ensure we ignore all the temp build files as we don't want to accidentally commit them to git

package.json
```diff
@@ -1,16 +1,18 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "5.5.0",
+  "version": "5.5.1-4",
```
dmbarry86, line 3: I've published a pre-release so I can test it in VIRTO.DATA.

package.json
mattcorner: https://app.warp.dev/block/98XN9HmSMLgfg51K0iW9zW

I tried installing this in VIRTO.TEST and it threw a bunch of errors in during build. 

Not sure how this worked originally because the old dependency was not a peerDependency. I suspect that we were getting lucky and one of our other common dependencies also relied on that polyfill so it technically was in our node_modules.

.eslintignore
mattcorner: Is this because we are compiling ts to js in place? Presumably once everything is in TS we can switch to using a glob for this?

mattcorner (changes requested):

dmbarry86:
It works fine in VIRTO.TEST. It's a peer dependency just like the old library was. Check my commit here https://github.com/IPG-Automotive-UK/virto/commit/9794b0315e159e17785cb9c56af9ca0fc6cf48aa

mattcorner:
> It works fine in VIRTO.TEST. It's a peer dependency just like the old library was. Check my commit here [IPG-Automotive-UK/virto@9794b03](https://github.com/IPG-Automotive-UK/virto/commit/9794b0315e159e17785cb9c56af9ca0fc6cf48aa)

I get that, but the react-ui package.json never lists its as a peerDependency, so I never got a warning.

dmbarry86:
> I get that, but the react-ui package.json never lists its as a peerDependency, so I never got a warning. Granted the old one didn't appear to be listed either.

I see. That wasn't clear from your first message. I can add it as (optional) peer dependency. I don't think you will get a warning though. It needs to be optional because not every component uses it.

mattcorner:
> > I get that, but the react-ui package.json never lists its as a peerDependency, so I never got a warning. Granted the old one didn't appear to be listed either.
> 
> I see. That wasn't clear from your first message. I can add it as (optional) peer dependency. I don't think you will get a warning though. It needs to be optional because not every component uses it.

In which case somethings not quite right either here, or in our Remix apps because VIRTO.TEST does not use any of the Card components that require this dependency, but they don't appear to be being tree-shaken out of the production builds. A problem for another time.

mattcorner (approved):

dmbarry86:
> In which case somethings not quite right either here, or in our Remix apps because VIRTO.TEST does not use any of the Card components that require this dependency, but they don't appear to be being tree-shaken out of the production builds. A problem for another time.

Yeah not sure about that. Not a new issue though I think because you had already installed the old dependency in VIRTO.TEST so must have come across the same. Looks like it's installed in every Remix app. Not sure what is special about this one that you had to install it but you didn't have to install every peer dependency you don't use.

PR #768, opened 2023-11-10
JS Declaration File Fix

dmbarry86:
Fixes issue with declaration files for JS components missing the prop type exports.


dmbarry86:
I've tested all changes here in a VIRTO branch for DATA, ID, VEHICLE and TEST. All apps build correctly and there are no TS warnings.

PR #769, opened 2023-11-13
Autocomplete Component - use isOptionEqualToValue method

syedsalehinipg:
## Changes

Added the use isOptionEqualToValue to fix the warning below:

image

This error started occurring after changes to the Autocomplete component was changed to also accept key value pairs as options.

With the use of the isOptionEqualToValue, we are now checking the value against the option based on the type of options being used:

- The function takes two parameters: `option` and `value`.
- It first checks if `option` is a `KeyValueOption` object using the `isKeyValueOption` type guard function. If `option` is a `KeyValueOption`, it compares `option.value` to `value`.
- If `option.value` is equal to `value`, it returns `true`, indicating that the option is equal to the value.
- If `option` is not a `KeyValueOption` (which means it's a string in this context), it compares `option` directly to `value`.
- If `option` is equal to `value`, it returns `true`, indicating that the option is equal to the value.
- If neither of the above conditions are met, it returns `false`, indicating that the option is not equal to the value.

This function is used to customize how options are compared to the current value in the Autocomplete component. By default, options are compared by reference, but this function allows options to be compared by their `value` property or their string value.

### DIFF
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -1,6 +1,5 @@
 import * as React from "react";
 
-import { AutocompleteProps, KeyValueOption } from "./Autocomplete.types";
 import {
   Box,
   Checkbox,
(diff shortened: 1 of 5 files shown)


m4manjesh (approved):
LGTM. The warning is not appearing now in key value options example.

PR #773, opened 2023-11-24
Migrating flaky tests to Playwright

VictorCLLucas:
## Changes

This change migrates the flaky unit tests that have been occurring [here](https://github.com/IPG-Automotive-UK/react-ui/actions/runs/6906451322/job/18791512769) to run using Playwright.
image

I removed the block of tests regarding the `onChange` checks under VehicleSelect. I recreated the tests using Playwright and placed them in a file in the same directory but in to a `.spec.tsx` file run when Playwright calls the tests. 

- Updated version of @playwright/test
- Recreated flaky tests in Playwright
- Removed flaky tests from the code

This PR is only to fix the flaky tests. More changes like this will come once a strategy for using Playwright in RUI has been implemented.

## Testing notes

- Checkout this branch
- Run `npm i -f`
- Run `npm run test` 
- Check that VehicleSelect tests for `onChange` are not ran under Unit tests
- Check that VehicleSelect tests are run under Playwright - this can be looked at by running `npx playwright show-report` after running `npx playwright test`


src/VehicleSelect/VehicleSelect.spec.tsx
```diff
@@ -0,0 +1,194 @@
...
+    ).toBeVisible();
+    await expect(
+      page.getByText('"64c8c4cccc8d6f00130b366b"').first()
+    ).toBeVisible();
+    await expect(page.getByText('"Gate 1"').first()).toBeVisible();
+    await expect(page.getByText('"Gate 1"').nth(1)).toBeVisible();
+    await expect(page.getByText('"Gate 2"').first()).toBeVisible();
+    await expect(page.getByText('"Gate 2"').nth(1)).toBeVisible();
+  });
+});
```
syedsalehinipg, line 194: I personally find the tests a bit difficult to follow, I would suggest maybe set up an arrays containing the value we expect for each option. And then loop with the array performing each check maybe. Maybe check with @dmbarry86 on his thoughts, below is an example..

```
import { expect, test } from "@playwright/test";

const selectors = [
  { label: "Project Code *", option: "911" },
  { label: "Model Year *", option: "2015" },
  { label: "Vehicle Variant *", option: "JS - 3.6 l6 - 397kW - 7MT -" },
  { label: "Vehicle Variant *", option: "MP - 3.6 l6 - 397kW - 7MT -" },
  { label: "Gate *", option: "Gate 1" },
  { label: "Gate *", option: "Gate 2" },
];

test.describe("Vehicle Select tests", () => {
  test("Test onChange is called with the expected values", async ({ page }) => {
    await page.goto(
      "http://localhost:6006/?path=/story/selectors-vehicleselect--default"
    );

    for (let i = 0; i < selectors.length; i++) {
      const { label, option } = selectors[i];

      // Open dropdown and select option
      await page
        .frameLocator('iframe[title="storybook-preview-iframe"]')
        .getByLabel(label)
        .click();
      await page
        .frameLocator('iframe[title="storybook-preview-iframe"]')
        .getByRole("option", { name: option })
        .click();

      // Check onChange calls
      await page.getByRole("tab", { name: "Actions 1" }).click();
      await expect(page.getByText("onChange")).toBeVisible();
      await expect(page.getByText(`onChange ${i + 1}`)).toBeVisible();
      await expect(page.getByText(`onChange ${i + 2}`)).not.toBeVisible();

      // Check onChange value
      await page.getByText("onChange").click();
      await page.getByText(i.toString()).click();
      await expect(page.getByText(`"${option}"`)).toBeVisible();
    }
  });
});
```

dmbarry86: I agree, it's very hard to follow. Also not helped by the fact there are no newlines between each code block so it's just one big blur. The suggestion from Syed is much easier to read. There's a tradeoff here between what is automatically generated by the Playwright test generator (presumably what Victor used) vs the optimised readable code a developer would write.

VictorCLLucas: Thanks for the feedback, I have addressed this and have submitted some changes. 
I based by implementation off your suggestion and made some modifications to it and the tests are passing, however I must highlight per each of the loops that it is doing it checks that the last value was added only, not that the other values remain.

VictorCLLucas:
Is this going to be the new flaky test 😅 
image

src/VehicleSelect/VehicleSelect.spec.tsx
syedsalehinipg: I would suggest that we also add a check after the for loop to check at the selected value in those field are as expected.

VictorCLLucas: I've added this

syedsalehinipg (approved):

PR #801, opened 2023-12-07
Removing dependency install from postinstall

VictorCLLucas:
## Changes

This change is to attempt to reduce the build times of VIRTO apps.

- renamed the `postinstall` script so that it doesn't download these dependencies every time `npm run install` is ran
- called the new script `test-deps` in `npm run test:playwright` so it only runs when running the tests


m4manjesh (approved):
LGTM. Playwright not called `npm run install` to install browsers

PR #808, opened 2024-01-25
TreeViewList - expand search items

syedsalehinipg:
## Changes

This PR introduces the functionality which expands out the items in the tree which matches the search term.

- A new flag has been added `expandSearchTerm` which can be used to control the use the functionality e.g when false items would not be expanded on search
- When enabled the `searchTerm` is used to search through a tree structure and return an array of node IDs that match a given search term. This allows users to easily find and navigate to items in the tree.
- The search is case-insensitive, meaning it will match nodes regardless of whether the search term is in upper case, lower case, or a mix.
- If a child node matches the search term, the function ensures that its parent node is also included in the result so that the parent is also visible. 

## UI/UX

image

image

## Testing notes

You can test by using the searchTerm input field in the controls of storybook.

_Note: expandSearchTerm to control usage and also a reload is required when the flag is changed_

### DIFF
--- a/src/TreeViewList/TreeViewList.tsx
+++ b/src/TreeViewList/TreeViewList.tsx
@@ -121,49 +121,39 @@ const TreeViewList = <T,>({
 
   // recursive search function
   const applySearch = (search: string, parameters: Item<T>[]): Item<T>[] => {
-    const terms = search
-      .toUpperCase()
-      .trim()
-      .split(/(?:\.| )+/);
+    // Split the search string into individual terms for comparison
+    const terms = search.toLowerCase().split(" ");
 
-    const searchInItem = (item: Item<T>): Item<T> | null => {
-      const match = terms.some(term =>
-        typeof item.name === "string"
-          ? item.name.toUpperCase().includes(term)
-          : false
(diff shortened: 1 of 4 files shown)


src/TreeViewList/TreeViewList.tsx
```diff
@@ -152,6 +179,21 @@ const TreeViewList = <T,>({
...
+      // Reset expanded ids
+      setExpanded([]);
+
+      // if the search term is not empty, get the ids of the matching nodes and set them as expanded
+      if (searchTerm !== "") {
+        const ids = searchTree(parameters, searchTerm);
+        setExpanded(ids);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
m4manjesh, line 194: We should not add this to avoid a dependency warning instead you can wrap `searchTree` in a useCallBack hook to fix it properly.

m4manjesh (changes requested):
Functionality looks good , a small suggestion to avoid disabling warnings.

m4manjesh (approved):
LGTM

PR #809, opened 2024-01-29
TreeViewList searching bug and simplified interface

syedsalehinipg:
Contributes to [TD-2030](https://sce.myjetbrains.com/youtrack/issue/TD-2030/Migrating-VIRTO.DATA-to-Remix)

## Changes

This PR fixes a bug with the searching logic in the TreeViewList component introduced in [v5.12.0](https://github.com/IPG-Automotive-UK/react-ui/releases/tag/v5.12.0) and also simplifies and tidies up the component interface to closer align with the core MUI components that are being wrapped.

* Previously the component accepted a searchTerm string but didn't actually provide the search box. This then meant that each consuming component would need to implement the same search box and pass along this prop. All of our current usecases include the search so we have decided to now include the search component directly in this component. It's optional and the default is not to show the search box. The search box is enabled with the enableSearch prop.
* The previously named `expandSearchTerm` prop has been renamed to `expandSearchResults` to provide clarity over the functionality. This can be set to true to expand all nodes in the tree when a search term is active. Default is false. This can be useful a UX improvement to help users quickly narrow down to an item without having to manually expand nodes.
* The `defaultExpanded` prop has been removed. We don't currently have a usecase for this.
* The `onSelectionChange` prop has been renamed to `onNodeSelect` to directly match the underlying MUI prop.
* A new `onNodeToggle` prop has been added so consumers can listen to expand/collapse events as well as selection events. VIRTO.BUILD will need this.
* The `items` prop has been re-typed and simplified a lot. You now pass `label` and `nodeId` which directly align with the underlying MUI props. You can provide optional props for `disabled` to disable a node and `tooltip` to show a tooltip on hover.
* Simplified logic that builds the tree nodes and fixed typescript issues. Previosuly we were trying to pass a generic but then making assumptions in the code based on specific props which could have lead to errors. The items prop is now explicitly typed as describe aboved.
* The tooltip component has been moved to a standalone component rather than a nested component to avoid re-renders.
* Fixed the logic that takes a search term or terms and filters the list of nodes displayed in the tree. It was previously displaying nodes that didn't match the search. Multiple search terms can be separated by a space. Each term must match a node or one of it's children for it to be displayed in the tree. Multiple terms are logical AND, i.e. all terms must match.
* Added a new story to storybook to show the optional tree with search.
* Fixed an issue where the width prop wasn't being used correctly. The width was being set but the component was still allowed to overflow. Now the tree and search (when shown) are wrapped in a container that correctly sets the width. Defeault is 100%.
* Renamed and shortened the example data used to populate the items in storybook. Previously we were passing every item from DATA which was hard to maintain. The new dataset provides enough variety to showcase all functionality and allow searching on multiple terms to filter down.

## UI/UX

Tree component with optional search.
image

image

Searching for single search term
image

Searching for multiple search terms to further narrow down
image


dmbarry86:
Handing this back over to @syedsalehinipg to tidy up the tests and PR description etc. It should be reviewed by somebody else as we have now both contributed.

src/TreeViewList/TreeViewList.tsx
```diff
@@ -1,335 +1,216 @@
...
- * Builds a tree structure from a flat list of items.
- *
- * @template T The type of the options array elements.
- * @param data - The flat list of items to transform into a tree.
- * @returns The tree structure built from the input items.
- * @example buildTree(items) // returns TreeNode[]
- */
-function buildTree<T>(data: Item<T>[]): TreeNode[] {
-  const itemsTree: TreeNode[] = [];
+// function to filter the items to contain only those that match the search term
```
Skoob1905, line 150: just expand the docs here to describe what the params are
```suggestion
/**
 * Filter Items to contain only those that match search
 * @param blah blah
 */
```

Skoob1905 (changes requested):
Just one small doc change but other than that, looks pretty good.

src/TreeViewList/TreeViewList.tsx
syedsalehinipg: Since this is a suggested code change it's best to have all the changes you suggest in the code, since you put 'blah blah' I can't commit this 😆

Skoob1905 (approved):

PR #811, opened 2024-02-03
Future/pnpm

mattcorner:
## Changes

Changes from using `npm` as a package manager to [`pnpm`](https://pnpm.io/).

Benefits:
- Code does not have access to dependencies of dependencies. The package.json must explicitly list all dependencies used in code.
- More disk efficient. Symlinks node_modules to a global store. 
- More performant. Keeps a global store so if a dependency already exists, it does not need to redownload.
- Github Action caching - Keep a global store for github actions.

More detail on Github Action Caching:
The cache key is `${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}`. 

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB. Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the oldest caches in the repository.

Hence as long as no pnpm-lock.yaml file have not changed, then we will hit the cache.

## Testing notes

All developers will need to install `pnpm`. There are many methods to doing so: https://pnpm.io/installation

- [x] ~Appropriate tests have been added.~


dmbarry86:
@mattcorner would you like me to have a bash with this?

mattcorner:
Couple more bits to wrap my head around, but i'll send it you way when ready.

dmbarry86:
I assume these are dependencies of dependencies and not something we can control?

image

dmbarry86 (approved):
LGTM. Let's just have a quick chat tomorrow before merging to align on couple of things.

mattcorner:
> I assume these are dependencies of dependencies and not something we can control?
> 
> image

Dependencies of plotly.js iirc.

dmbarry86:
I think we're good to go with this now @mattcorner? I assume your prerelease worked as expected?

PR #815, opened 2024-02-16
Enhancement - TreeViewList improvements

syedsalehinipg:
## Changes

This PR includes the following changes:

- onNodeSelect function to pass back node detail e.g. if the selected node is a leaf 
- **Bug-** multiple tooltips was showing parent and child nodes together
  - fixed to show only one at a time, for both parent and child, also fixed position to be RIGHT-START
- Fixed a performance issue when there are a large number of items the search was lagging behind by few seconds inconsistently, to deal with this a new prop is has been added _expandItem_  this is the number of items to expand, it is defaulted to 1. Also with the use of debounce the performance has significantly improved.
- Fixed an issue where the user expanded nodes would close when using search and then clearing the search

Other:
  - Fixed a failing tests using `{ exact: true }` to match the value when using `getByText`
  - Added new `expandItems` prop to a URL param
  - Changed element count to match expected results

### onNodeSelect:
_Here when the parent was clicked, a object was returned with a **isChild** property as **false** and then when the child was clicked a object was returned with a **isChild** property as **true**._
image

### Expanding matching item:
_Here **expandSearchResults** is set to true and **expandItems** is set to 2, which is showing 2 leaf nodes as expected._
image

### Tooltip showing correctly:
_Tooltip is displayed for parent node._
image

_Tooltip is displayed for child note._
image

## Testing notes

**Large dataset:**
Download this: [data.json](https://github.com/IPG-Automotive-UK/react-ui/files/14418286/data.json)
Replace contents in: _**example-items.json**_

**onNodeSelect function to pass back if the selected node is a leaf**
 - See the actions tab for the third value on OnNodeSelect

**Bug -multiple tooltips was showing parent and child nodes together**
 - Hover over nodes, you should only see one tooltip open at once.

**Fixed a performance issue when there are a large number of items the search was lagging behind by few seconds...**
 - set the number of expandItems and perform a search (note: if set to high number e.g. 200, there will be a lag and maybe flicker, however that is not a realistic use case)

*Fixed an issue where the user expanded nodes would close when using search and then clearing the search**
1. Before searching, expand out a node.
2. Perform a search that does not include the already expanded node.
3. Then clear the search field.
4. You should see the one you expanded before the search, still expanded.


src/TreeViewList/TreeViewList.tsx
```diff
@@ -57,24 +59,32 @@ const TreeViewList = ({
     }
   }, [enableSearch, items, searchValue]);
 
-  // if search is enabled and expandSearchResults is true, expand all nodes when the tree display items change
-  useEffect(() => {
-    if (enableSearch && expandSearchResults && searchValue !== "") {
-      const expandedNodes: string[] = [];
-      const expandAllNodes = (items: TreeNodeItem[]) => {
+  const expandNodes = () => {
```
mattcorner, line 62: Missing code doc

mattcorner (changes requested):

src/TreeViewList/TreeViewList.tsx
```diff
@@ -130,10 +204,21 @@ const TooltipTreeItem = (
...
     <Tooltip
       disableFocusListener
-      title={props.tooltip ? <>{props.tooltip}</> : ""}
+      title={
+        props.tooltip && !hasChildren(props.children) ? (
+          <>{props.tooltip}</>
+        ) : (
+          ""
+        )
+      }
```
dmbarry86, line 221: Not sure I agree with this change. If you only want tooltip for children in a specific application then don't provide tooltip prop for the parents. Don't think this logic should live in the component as it already has an optional interface to set tooltip on each node.

src/TreeViewList/TreeViewList.tsx
```diff
@@ -115,6 +165,30 @@ const TreeViewList = ({
...
+    } else if (node.children) {
+      const result = getNodeById(node.children, nodeId);
+      if (result) {
+        return result;
+      }
+    }
+  }
+  return null;
+};
+
```
dmbarry86, line 191: Is it the most performant way to be calling this recursive function every time a selection is made? We already map through all of the nodes on initial render to build the tree so why not just store a map of child node ids once then? I imagine indexing into a map is better than calling this function each time.

dmbarry86: I suppose it's also questionable whether this functionality should live in the component itself or whether the consuming service just deals with it. If the consuming service only cares about node selection for certain nodes (child nodes being one such example) then just store a map yourself when you build the tree data structure and use that in your own onNodeSelect handler.

dmbarry86: We profiled this together and with the largest dataset we have (DATA) it seems to only be 2-5ms with this implementation so we keep it for now.

dmbarry86:
On the call together we noted the following:

* Flickering when removing a search term.
* The nodes that the user manually expanded get lost when a search is entered. The old tree in DATA remembered the user expanded nodes so we should probably move to do the same. Instead of controlling defaultExpanded we need to setting expanded on the tree and then being on control of tracking the state when any node is expanded. This should work together with the option to auto expand nodes when the prop is set.

dmbarry86 (changes requested):
I don't think the logic to return to the previous user selected expansion state when clearing search is working.

syedsalehinipg:
https://github.com/IPG-Automotive-UK/react-ui/assets/143519265/8e8df79b-340e-49c5-bb37-e31d4b767739

PR #826, opened 2024-03-05
TD-1854 Open current app link in the current tab

dmbarry86:
Contributes to [TD-1854](https://sce.myjetbrains.com/youtrack/issue/TD-1854/Suppress-new-VIRTO-tab-opening-when-switching-apps)

## Changes

Updated the AppLauncher component so that links to the current app I am in open in the current tab and new apps open in a new tab. Previously all links opened in a new tab but it was requested from @Sowbhagya-ipg that links to the current app should just do a page reload in the current window.

For example here I am currently in SCENE so clicking the SCENE app menu item should reload in the current window.

We did also discuss the idea of removing links for the current app I am in but it was decided instead to keep showing them but reload in the current window rather than new window.

image

## Dependencies

N/A

## UI/UX

N/A

## Testing notes

I've released a pre-release and integrated it in to VIRTO as that's the easiest way to test this behaviour which relies on the URL. You can use the deploy preview [here](http://k8s-deploypr-apigatew-4ef71a0284-622295924.eu-west-2.elb.amazonaws.com/) to test the behaviour in one of the apps. For example, open VIRTO.SCENE then use the app launcher component and click on SCENE. You should get a page refresh in the current tab. Then click another app and it should open in a new window.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "7.0.3-0",
+  "version": "7.0.3-1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ipguk/react-ui",
-      "version": "7.0.3-0",
+      "version": "7.0.3-1",
       "license": "MIT",
       "dependencies": {
         "@juggle/resize-observer": "^3.4.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "7.0.2",
+  "version": "7.0.3-0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ipguk/react-ui",
-      "version": "7.0.2",
+      "version": "7.0.3-0",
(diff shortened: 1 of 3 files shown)


m4manjesh:
@dmbarry86  If i am in BUILD and try to access FLEET via AppLauncher FLEET load on same tab but if i am in  FLEET and try to aceess BUILD it open in a new tab. Is that the expectd behaviour for FLEET/BUILD? Also TEST link seems broken so i couldnt test it.
apps-links

dmbarry86:
That's because of the weirdness in FLEET/BUILD where we treat them as two apps but served from the same app. They both share the same basename so fleet is /fleet/ and build is /fleet/build. I don't see any way around that with the current routing.

TEST issue is because deploy preview ran out of nodes. Will be fine.

m4manjesh:
> @m4manjesh That's because of the weirdness in FLEET/BUILD where we treat them as two apps but served from the same app. They both share the same basename so fleet is /fleet/ and build is /fleet/build. I don't see any way around that with the current routing.
> 
> TEST issue is because deploy preview ran out of nodes. Will be fine.

Okay, then changes looks good to me.

m4manjesh (approved):

PR #835, opened 2024-03-22
Uncontrolled input  component updates

lukemojo:
Closes TD-2588

## Changes

Add props to RUI components to make inputs optionally controlled or uncontrolled.

This pull request will allow us to use each input in an uncontrolled manner for new validated form components. The name field allows the validation library to manage the input fields.

- Add defaultValues/defaultChecked and a name prop to the following components
- Checkbox
- RadioGroup
- TextField
- Switch
- Autocomplete

## UI/UX

No changes

## Testing notes

View the uncontrolled stories in storybook. All components should function as normal and the initial render should render the defaultValue.

### DIFF
--- a/src/Autocomplete/Autocomplete.stories.tsx
+++ b/src/Autocomplete/Autocomplete.stories.tsx
@@ -60,20 +60,6 @@ const Template: StoryFn<
   );
 };
 
-const UncontrolledTemplate: StoryFn<
-  AutocompleteProps<string | KeyValueOption, boolean | undefined>
-> = args => {
-  // Return the Autocomplete component with the appropriate props
-  return (
-    <Autocomplete
-      {...args}
-      onChange={(event, newValue) => {
-        action("onChange")(newValue);
-      }}
-    />
-  );
-};
-
 // Define the default story
 export const Default: StoryObj<typeof Autocomplete> = {
   args: {
@@ -100,10 +86,11 @@ export const Default: StoryObj<typeof Autocomplete> = {
   },
   render: Template
 };
-// Define the default story
+
+// Define the uncontrolled story
 export const Uncontrolled: StoryObj<typeof Autocomplete> = {
   args: {
-    // Define the default args
+    // Define the uncontrolled args
     defaultValue: "Option 4",
     disabled: false,
     error: false,
@@ -124,7 +111,7 @@ export const Uncontrolled: StoryObj<typeof Autocomplete> = {
     size: "medium",
     variant: "outlined"
   },
-  render: UncontrolledTemplate
+  render: Autocomplete
 };
 
 // Define the story for key-value options
--- a/src/Autocomplete/Autocomplete.stories.tsx
+++ b/src/Autocomplete/Autocomplete.stories.tsx
@@ -60,6 +60,20 @@ const Template: StoryFn<
   );
 };
 
+const UncontrolledTemplate: StoryFn<
+  AutocompleteProps<string | KeyValueOption, boolean | undefined>
+> = args => {
+  // Return the Autocomplete component with the appropriate props
+  return (
+    <Autocomplete
+      {...args}
+      onChange={(event, newValue) => {
+        action("onChange")(newValue);
+      }}
+    />
+  );
+};
+
 // Define the default story
 export const Default: StoryObj<typeof Autocomplete> = {
   args: {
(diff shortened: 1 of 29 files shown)


src/SwitchField/index.ts
```diff
@@ -3,9 +3,11 @@ import type { SwitchProps } from "@mui/material";
 
 export type SwitchFieldProps = {
```
Skoob1905, line 4: add some docs here

src/RadioButtons/index.ts
```diff
@@ -3,8 +3,10 @@ import type { RadioGroupProps, SxProps, Theme } from "@mui/material";
 import RadioButtons from "./RadioButtons";
 
 export type RadioButtonsProps = {
+  defaultValue?: string;
```
Skoob1905, line 6: add couple docs here

src/Checkbox/index.ts
```diff
@@ -2,8 +2,10 @@ import Checkbox from "./Checkbox";
 
 export type CheckboxProps = {
   checked?: boolean;
+  defaultChecked?: boolean;
   disabled?: boolean;
   label?: string;
+  name?: string;
```
Skoob1905, line 8: add some docs here

Skoob1905 (changes requested):

src/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -86,6 +100,32 @@ export const Default: StoryObj<typeof Autocomplete> = {
   },
   render: Template
 };
+// Define the default story
```
mattcorner, line 103:
```suggestion
// Define the uncontrolled story
```

src/SwitchField/SwitchField.test.jsx
```diff
@@ -18,6 +18,19 @@ describe("SwitchField", () => {
...
+    const user = userEvent.setup();
+    const onChange = vi.fn(event => event.target.checked);
+    render(
+      <SwitchField
+        defaultChecked={true}
+        options={["Option A", "Option B"]}
+        onChange={onChange}
+      />
+    );
+    await user.click(screen.getByRole("checkbox"));
```
mattcorner, line 31: Your tests seem to either check the rendered value given a defaultValue, or in this case check the onChange handler after a user interaction. I suggest we need both for all the touched components in this PR.

src/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -86,6 +100,32 @@ export const Default: StoryObj<typeof Autocomplete> = {
...
+      "Option 3",
+      "Option 4",
+      "Option 5",
+      "Option 6"
+    ],
+    required: false,
+    size: "medium",
+    variant: "outlined"
+  },
+  render: UncontrolledTemplate
```
mattcorner, line 127: Do you need the UncontrolledTemplate wrapper in all these stories, or can you just pass the component in directly? It just seems we are removing all state etc which ends up just being the underlying component.

The TextField story is a clear example of this.

```suggestion
  render: Autocomplete
```

mattcorner:
image

I know its not related to this PR, but I saw this in the storybook logs. Can we just delete the right hand side of this code please? We never hit it as vitejs tells us.

mattcorner (changes requested):
Generally looks good. 

Up to you, but it might be a good idea to do a prerelease of RUI from this branch and install it in a couple apps. We're changing some pretty fundamental components here so just to make sure we haven't broken anything.

src/SwitchField/SwitchField.test.jsx
lukemojo: In this instance I was only checking the onChange to infer the default value but have updated it to just check the default value like the checkbox. The onChange events for each component haven't been touched and will just reflect the change event  as per existing behaviour.

Skoob1905 (approved):
- **Autocomplete:** I uploaded a carmaker test run both to local and devel and then viewed an autocomplete component behaviour is the same.
- **Checkbox:** I changed the road filters in general and the behaviour is matched.
- **LabelSelector:** I uploaded a road with a label called test and I can filter in both devel and locally
- **Radio:** Formgroups are conditionally rendered when using radio buttons, behaviour is the same.
- **Switchfield:** I can successfully change between admin and standard in VIRTO.ID and 
- **Textfield:** Tested this using the search filter, Germany yielded some results and "Germanu" returned no results found for both local and devel

These changes haven't affected any current vital functionality in the app. The usage of using the defaultValues needs to be tested at a later date as I don't have the "middleware" yet to convert the map into a component, but this will come in a later PR.

Skoob1905:
Probably best to just get Matts seal of approval on this before merging this in.

lukemojo:
Please note I've now added a name prop to autocomplete

mattcorner:
Conflict to resolve, then go ahead and merge 👍

mattcorner (changes requested):

mattcorner (approved):
lgtm

PR #841, opened 2024-04-02
TD-2594 Create React UI - Variant vehicle search component

syedsalehinipg:
Closes - [TD-2594](https://sce.myjetbrains.com/youtrack/issue/TD-2594)
Contributes to - [TD-2030](https://sce.myjetbrains.com/youtrack/issue/TD-2030)

## Changes

This PR introduces a new VehicleSelector component. This component is used to select a vehicle based on various parameters such as project code, model year, vehicle variant, and gate. 

_This component is similar to VehicleSelector but Gate is not require here so it is made optional and also we are not using multiselect on all fields which removes the complexity._ 

Key features:

- Uses the Autocomplete component for selection.
- Allows for the selection of project code, model year, vehicle variant, and gate.
- Handles disabled states based on the selection of previous parameters.
- If no gates are passed in, the Gate field will be hidden.
- If `multipleSelection` is set to `true`, Gate field will allow multiple selection.

Storybook Stories:

`Default`: The default configuration of the `VehicleSelector `component.
`WithGate`: The `VehicleSelector `component with gate selection enabled.
`MultiselectWithGate`: The `VehicleSelector `component with gate selection and multiple selection enabled.

## UI/UX

_flexDirection - column_
image

_flexDirection - column - dark mode_
image

_flexDirection - row_
image

_flexDirection - row dark mode_
image

## Testing notes

Run `pnpm run storybook` to view and interact with the `VehicleSelector` component in isolation. Use the required props such as flexDirection, flexWrap, gates, multipleSelection, onChange, value, and variants.

~~- [ ] Branch has been run in docker.~~

### DIFF
--- a/src/VehicleSelector/VehicleSelector.spec.tsx
+++ b/src/VehicleSelector/VehicleSelector.spec.tsx
@@ -135,3 +135,78 @@ test("Autocomplete components are enabled or disabled based on the selection of 
       .getByLabel("Model Year *")
   ).toBeEnabled();
 });
+
+/**
+ * Test to check that clearing the "Project Code" field also clears the "Model Year", "Vehicle Variant", and "Gate" fields.
+ */
+test("Clearing the 'Project Code' field also clears the 'Model Year', 'Vehicle Variant', and 'Gate' fields", async ({
+  page
+}) => {
+  // Go to the page with the VehicleSelector component
+  await page.goto(
+    "http://localhost:6006/?path=/story/selectors-vehicleselector--with-gate"
+  );
+
+  // Wait for the iframe to be attached in the DOM.
+  await page.waitForSelector('iframe[title="storybook-preview-iframe"]');
+
+  // Select a value from each Autocomplete component
+  await page
+    .frameLocator('iframe[title="storybook-preview-iframe"]')
+    .getByLabel("Project Code *")
+    .click();
+  await page
+    .frameLocator('iframe[title="storybook-preview-iframe"]')
+    .getByRole("option", { name: "911" })
+    .click();
+  await page
+    .frameLocator('iframe[title="storybook-preview-iframe"]')
+    .getByLabel("Model Year *")
+    .click();
+  await page
+    .frameLocator('iframe[title="storybook-preview-iframe"]')
+    .getByRole("option", { name: "2015" })
+    .click();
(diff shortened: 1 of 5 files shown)


src/VehicleSelector/VehicleSelector.tsx
```diff
@@ -0,0 +1,247 @@
...
+              }
+            ]);
+          }}
+          value={selectedModelYear === "" ? null : selectedModelYear}
+        />
+      </Box>
+      <Box flex="40%">
+        <Autocomplete
+          disabled={selectedModelYear === null || selectedModelYear === ""}
+          label="Vehicle Variant"
```
arpadlomenipg, line 126: I am not sure what part causes this, but there are some issues with the "Vehicle Variant" field, when field values are deleted, see the details below.

When deleting the content of the "Project code" field, every other filed should be emptied. When deleting the content of the "Model Year" filed, every field other than "Project Code" should also be deleted. (As seen in component "VehicleSelect")
However, when these two actions are done, the "Vehicle Variant" field still retains its value (albeit not editable anymore). Which can lead to a scenario, when after deleting e.g. a "Project Code", then adding a new "Project Code", an invalid "Vehicle Variant" will be selected, since it retained its value after deleting the first "Project Code".

Another issue arises when Gates are enabled, and the content of the "Vehicle Variant" is deleted. In this case, the contents of "Gates" should be emptied. But the "Gates" field retains its value in this scenario.

src/VehicleSelector/VehicleSelector.spec.tsx
```diff
@@ -0,0 +1,137 @@
...
+    .getByRole("option", { name: "911" })
+    .click();
+
+  // Check that "Model Year" is now enabled
+  await expect(
+    page
+      .frameLocator('iframe[title="storybook-preview-iframe"]')
+      .getByLabel("Model Year *")
+  ).toBeEnabled();
+});
```
arpadlomenipg, line 137: I would also add a test, where e.g. the contents of "Project Code" are deleted, then check if all other fields' contents are also deleted, as they should be.

arpadlomenipg (changes requested):

arpadlomenipg (approved):
LGTM

PR #844, opened 2024-04-18
TD 2659 Update NumberField and FileUploaders

jegasriskantha:
contributes to [TD-2659](https://sce.myjetbrains.com/youtrack/issue/TD-2659/Create-Step-2)

## Changes

 - Updates NumberField to ensure value remains undefined if no value is passed, which also enables default value to be passed into the component. 
 - Updates the ImageUploader and FileUploader to optional accept an id as an input, and updates the onDelete callback to return the deleted file as a second argument. This was done so that when a file uploader accepts multiple files we can accurately determine which file needs to be deleted by it's id.  Where this updated component has been implemented, we are saving the file in a session where the id references a temporary folder where the file is stored. When the user deletes the file, the the deleted file (and it's id) is send to the server so that we can delete only the file that matches that id from the session. 

## Dependencies

N/A

## UI/UX

No real UI updates other than NumberField can now accept a default value. 

## Testing notes

Automated Testing: 
- Added test to NumberField to confirm that default value is rendered when default value is passed in. 

### DIFF
--- a/src/NumberField/NumberField.jsx
+++ b/src/NumberField/NumberField.jsx
@@ -26,7 +26,7 @@ export default function NumberField({
   startAdornment = null,
   step,
   stepper = true,
-  value = null,
+  value,
   variant = "outlined"
 }) {
   // state to keep track of the value of the number field even when invalid
--- a/src/NumberField/NumberField.jsx
+++ b/src/NumberField/NumberField.jsx
@@ -99,7 +99,7 @@ export default function NumberField({
       required={required}
       size={size}
       type="number"
-      value={currentValue ?? ""}
+      value={currentValue}
       variant={variant}
       InputProps={{
         ...(endAdornment && {
--- a/src/NumberField/NumberField.stories.jsx
+++ b/src/NumberField/NumberField.stories.jsx
@@ -43,6 +43,7 @@ export const Default = {
 
 export const Uncontrolled = {
   args: {
+    defaultValue: 12,
     disabled: false,
     error: false,
     helperText: "What Number are you going to type?",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "7.4.0",
+  "version": "7.4.1-0",
   "description": "React UI component library for IPG web applications",
   "author": "IPG-Automotive-UK",
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ipguk/react-ui",
-  "version": "7.4.1-1",
+  "version": "7.4.1-2",
   "description": "React UI component library for IPG web applications",
   "author": "IPG-Automotive-UK",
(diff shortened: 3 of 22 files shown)


src/NumberField/NumberField.stories.jsx
syedsalehinipg, line 1: When I update the default value using the prop, I don't see the value reflected in the field. 

Here I have change the value to 5 but the field is still showing as 12:

image

src/NumberField/NumberField.jsx
syedsalehinipg, line 1: If a `min` and `max` are set, but the default value is not within that range, we should handle this?

Maybe something like this:

```
const [currentValue, setCurrentValue] = useState(() => {
  let initialVal = defaultValue || min;
  if (initialVal < min) initialVal = min;
  if (initialVal > max) initialVal = max;
  return initialVal;
});
```

In this code, the initial value is set to the `defaultValue` or `min` if `defaultValue` is not provided. Then, if the initial value is less than `min`, it's set to `min`. If it's greater than `max`, it's set to `max`. This ensures that the initial value is always within the specified range.

syedsalehinipg (changes requested):
I am unsure how to test the and verify the changes to the FileUploader and ImageUploader,

This is what I attempted:

1. Uploaded two files
2. Deleted one - I see in the actions one object with data and file
3. Deleted next one - the response is a blank array

Should I be seeing a object for each one, to identify which one was deleted?

image

Can I have some more instructions on how to test this please?

src/NumberField/NumberField.stories.jsx
jegasriskantha: This default value only works on first render, if you refresh the page you'll see that it has been updated.

src/NumberField/NumberField.jsx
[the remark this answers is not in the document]
jegasriskantha: No, the current value shouldn't be set at all until the user has interacted. Once the value is set it becomes a controlled component.

jegasriskantha: HTML standards don't validate on on default value, so I don't think we should either.

jegasriskantha:
> I am unsure how to test the and verify the changes to the FileUploader and ImageUploader,
> 
> This is what I attempted:
> 
> 1. Uploaded two files
> 2. Deleted one - I see in the actions one object with data and file
> 3. Deleted next one - the response is a blank array
> 
> Should I be seeing a object for each one, to identify which one was deleted?
> 
> image
> 
> Can I have some more instructions on how to test this please?

Sorry I didn't provide more info. 

I have update both the ImageUploader and FileUploader stories so that the onDelete action returns both arguments. You'll now see that when you delete a file you get both the current selection and the deleted file object.  This is the way to check that the second argument is the deleted object. 

I have also create an new story on FileUploader that passes in the optional id field, you should now see this id when when you delete the file. This is the way to check that the id is passed through the component.

syedsalehinipg (approved):
LGTM

PR #810, opened 2024-02-02
Future/vite

mattcorner:
Closes TD-??
Contributes to TD-???

## Changes

A brief description of what this pull request solves / introduces.

- Some detailed bullet points on what exactly was done.
- ...
- ...

Maybe a short description of what is not included and will come in future work where appropriate.

## Author checklist before assigning a reviewer

- [ ] Reviewed my own code-diff.
- [ ] Branch has been run in docker.
- [ ] PR assigned to me or an appropriate delegate.
- [ ] Relevant labels added to the PR.
- [ ] Appropriate tests have been added.
- [ ] Lint and test workflows pass.

PR #848, opened 2024-04-25
Create traffic sign component

AronPetrauskas:
## Changes

A new RUI component is introduced that renders traffic signs onto a konva canvas.

- TrafficSign component created.
- TrafficSign component variations created in storybook.
- End-to-end tests written for the TrafficSign component.

## UI/UX

Highway traffic sign example:
image

## Testing notes

> [!IMPORTANT]
> **Failed CI tests are a known issue, please ignore.** Tests failing because of "timeout" are a known flaky test issue.

1. Checkout the branch
2. `pnpm i`
3. `pnpm run storybook`
4. Test the 'TrafficSign' component under Road View (category)
Test all props can be changed and the component updates as expected.

image

- ~[ ] Branch has been run in docker.~


dmbarry86:
> 5\. Using the testing tab in VSCode run the tests manually. Make sure they all run. **_Make sure storybook is running before testing!_**
> 
> Note
> 
> **You'll have to run the test at least twice.** First run generates the reference images. Make sure the reference images display the expected component.

I don't think we should recommend manually running tests in PRs like this. You should recommend to run the tests via Docker, for which snapshots already exist. Don't confuse it for reviewers.

dmbarry86:
> A new RUI component is introduced that renders traffic signs onto a konva canvas.
> 
> * TrafficSign component created.
> * TrafficSign component variations created in storybook.
> * End-to-end tests written for the TrafficSign component.

It would be nice to show a screenshot of the component in future reviews. That's what the UI section is for (which you deleted by the look of it).

dmbarry86:
You need to make sure your component is exported from the root index otherwise you won't be able to import it in your app. Same goes for figure component which also appears to be missing although already merged.

https://github.com/IPG-Automotive-UK/react-ui/blob/create-traffic-sign-component/src/index.ts

src/TrafficSign/TrafficSign.types.ts
```diff
@@ -0,0 +1,126 @@
...
+    | "unevenRoad"
+    | "urbanArea"
+    | "urbanAreaEnd"
+    | "warningPedestrianCrossing"
+    | "weightLimit"
+    | "widthLimit";
+  /**
+   * The coordinates of the sign in Fr0 [x,y]
+   */
+  points: number[];
```
dmbarry86, line 102: Is this typed correctly? number[] allows any length array, e.g. [1, 2, 3, 4]. Are you after [1, 2] or [ [1,2], [3,4] ]?

```
type option1 = number[]
type option2 = [number, number]
type optionm3 = [number, number][]
```

src/TrafficSign/TrafficSign.types.ts
```diff
@@ -0,0 +1,126 @@
...
+   * The scale of the image
+   */
+  scale?: number;
+  /**
+   * The size of the image individually in x and y
+   */
+  size?: {
+    x: number;
+    y: number;
+  };
```
dmbarry86, line 121: Is this some pixel sizing? Width and height would be more traditional props one might expect if so.

src/TrafficSign/TrafficSign.types.ts
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * A function that resolves when the traffic sign is fully drawn to the canvas
+   */
+  onTrafficSignLoad?: (loaded: boolean) => void;
```
dmbarry86, line 125: I might called this onTrafficSignLoaded rather than onTrafficSignLoad if it runs after it has loaded rather than when it is about to load which the current name implies.

src/TrafficSign/TrafficSignHelper.ts
```diff
@@ -0,0 +1,347 @@
...
+import turnL from "./svg/turnL.svg";
+import turnR from "./svg/turnR.svg";
+import unevenRoad from "./svg/unevenRoad.svg";
+import urbanArea from "./svg/urbanArea.svg";
+import urbanAreaEnd from "./svg/urbanAreaEnd.svg";
+import warningPedestrianCrossing from "./svg/warningPedestrianCrossing.svg";
+import weightLimit from "./svg/weightLimit.svg";
+import widthLimit from "./svg/widthLimit.svg";
+
+interface IRoadSignsObject {
```
dmbarry86, line 159: type rather than interface should be the default unless you need any features of interface

dmbarry86, line 159: Add some comments

src/TrafficSign/TrafficSignHelper.ts
```diff
@@ -0,0 +1,347 @@
...
+  [key: string]: string;
+}
+
+const roadSigns: IRoadSignsObject = {
```
dmbarry86, line 163: add some comments

src/TrafficSign/TrafficSignHelper.ts
```diff
@@ -0,0 +1,347 @@
...
+    "minimumSpeedEnd",
+    "speedLimitZone",
+    "speedLimitZoneEnd"
+  ];
+  if (signsWithValues.some(signWithValue => type === signWithValue)) {
+    return roadSigns[`${type}${value}`] ?? fileUnknown;
+  } else {
+    return roadSigns[`${type}`] ?? fileUnknown;
+  }
+};
```
dmbarry86, line 345: What is stopping me from passing an unsupported value, e.g. type = speedLimit and value = 10000?

src/TrafficSign/TrafficSignHelper.ts
```diff
@@ -0,0 +1,347 @@
...
+    "speedLimit",
+    "speedLimitEnd",
+    "steepAscent",
+    "steepDescent",
+    "minimumSpeed",
+    "minimumSpeedEnd",
+    "speedLimitZone",
+    "speedLimitZoneEnd"
+  ];
+  if (signsWithValues.some(signWithValue => type === signWithValue)) {
```
dmbarry86, line 340: I find this option easier to read and I think is actually what you want to check rather than some

```suggestion
  if (signsWithValues.includes(type)) {
```

dmbarry86: There's also no need for the else here. We generally try to avoid if/else where the if returns or errors.

```
if (condition) {
  return something
}

return somethingElse
```

package-lock.json
```diff
@@ -18,16 +18,19 @@
         "@types/react-plotly.js": "^2.6.3",
         "@types/react-text-mask": "^5.4.14",
         "colord": "^2.9.3",
+        "konva": "^9.3.6",
```
dmbarry86, line 21: I'm curious why we have a package lock file again now in the repository and indeed why it changed here on this PR even though the package.json didn't change. When pnpm was introduced the package lock was removed and replaced by pnpm lock file. https://github.com/IPG-Automotive-UK/react-ui/pull/811

Maybe you can track down with @syedsalehinipg what has happened.

dmbarry86: I suspect maybe it was your figure PR that reintroduced it. Maybe you ran npm install rather than pnpm install?

dmbarry86:
The default story should show default values for all props. The "set number" buttons etc mean there isn't currently a value defined in the story. Make sure all stories have appropriate values defined for all props.

image

src/TrafficSign/TrafficSign.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A UI component that displays an SVG of the road sign on a canvas.
+ * @param angle The angle of rotation of the image in degrees
+ * @param type The name of the road sign
+ * @param value The value within the road sign (e.g., "50") for a speed limit
+ * @param points List of coordinates of the image on the canvas [x,y]
+ * @param scale A number to scale the image, keeping the aspect ratio
+ * @param size An object that allows sizing x and y independently
+ * @returns React Konva Image component
```
dmbarry86, line 14: No need to document all of props here when the input is an argument. We document for positional arguments but for something that takes an object just document the object type (TrafficSignProps) and then only document each prop in the type definition itself. Just duplicating effort otherwise.

src/TrafficSign/TrafficSign.types.ts
```diff
@@ -0,0 +1,126 @@
...
+   */
+  value?: string;
+  /**
+   * A rotation angle in degrees
+   */
+  angle?: number;
+  /**
+   * The scale of the image
+   */
+  scale?: number;
```
dmbarry86, line 114: What unit is this in? I entered a value of -100 and something rendered but I can't quite work out what the scale is actually doing.

image

dmbarry86: Is negative somehow flipping as well? I assume this is allowed/expected syntax?

src/TrafficSign/TrafficSign.types.ts
```diff
@@ -0,0 +1,126 @@
...
+    | "weightLimit"
+    | "widthLimit";
+  /**
+   * The coordinates of the sign in Fr0 [x,y]
+   */
+  points: number[];
+  /**
+   * The value of the sign (e.g., speed limit)
+   */
+  value?: string;
```
dmbarry86, line 106: The combination of type and value seems weird to me. Also means I can enter unsupported values. Why not just allow speedLimit100, speedLimit60 etc as types and do away with value? The list becomes longer of course but it's already very long and you would then get proper type support.

image

dmbarry86:
In future, just delete sections that are no longer relevant rather than strikethrough. I don't need to see the past state when I'm reviewing it today.

image

dmbarry86 (changes requested):
I've had an initial run through and bash. Some general and some specific comments. I've told the team to hold off reviewing the other PRs for now as some of the general points here will also come up in those PRs so it seems best to get one right first and then ask people to review the others in which case hopefully they are then just picking up on specific things for that component. I appreciate some of the general points are because you are new to the team. I've asked the senior developers to start a code style guideline document so maybe you could also contribute to that with the things you have picked up so far with your time in the team which we don't yet have written down. @jegasriskantha is going to start that doc.

src/TrafficSign/TrafficSign.types.ts
AronPetrauskas: Typed wrong, intended was `[number. number]`

src/TrafficSign/TrafficSign.types.ts
AronPetrauskas: It is pixel sizing to my understanding, however it is a bit more complex than that. Giving this prop a width 10 will not mean that the image necessarily will be 10px wide on the canvas. The stage/canvas itself has a scale prop which will scale the pixels, same goes for the scale prop of the TrafficSign component. Anyhow, the prop naming has been changed to `width` and `height`

src/TrafficSign/TrafficSign.types.ts
AronPetrauskas: Makes sense, changed.

src/TrafficSign/TrafficSignHelper.ts
AronPetrauskas: Fixed

src/TrafficSign/TrafficSign.types.ts
AronPetrauskas: Yeah, so this is a factor that scales the size of image/traffic sign. The native behavior of the library is that a negative value will 'flip' the image. I can of course disable this and only allow positive numbers, but my logic was to keep it as close to the native thing as possible. For now I have added a comment in the types, hopefully explaining it better.

image

src/TrafficSign/TrafficSign.tsx
AronPetrauskas: Should these be the defaults of the component? For example the default value for `scale` of trafficSign is `undefined`. This is so because the underlying component `Figure` will default `scale` to 3. So should the default value for `scale` for `TrafficSign` be `undefined` or `3` for storybook?

src/TrafficSign/TrafficSign.types.ts
AronPetrauskas: After discussion, the value field has been removed. All signs are accessible only via the type string. In the future if support for custom text is added, it will be done via a `label` prop.

src/TrafficSign/TrafficSign.tsx
AronPetrauskas: No defaults set to `undefined`

package-lock.json
AronPetrauskas: Probably was an `npm i` that caused this. `package-lock.json` removed

dmbarry86:
I don't think this is related to your changes but there's somethign weird with storybook giving me this error for your component and all components. If I open in incognito browser it works fine but doesn't work in my usual browser window. Some weird cache or something maybe between branches. I'm copying in @mattcorner to take a look as I wonder if it's related to vite changes https://github.com/IPG-Automotive-UK/react-ui/pull/810.

image

No need to hold up this PR with it.

PR #849, opened 2024-04-25
TD-2675 Create roadview road components

AronPetrauskas:
## Changes

Two new RUI components are introduced that render the road surface and road markings on to a konva canvas.

- RoadSurface and RoadMarking component created.
- RoadSurface and RoadMarking component variations created in storybook.
- End-to-end tests written for the RoadSurface and RoadMarking components.

### RoadMarking:
image

### RoadSurface:
image

## Testing notes

> [!IMPORTANT]
> **Failed CI tests are a known issue, please ignore.**

1. Checkout the branch
2. `pnpm i`
3. `pnpm run storybook`
4. Test the 'RoadSurface' and 'RoadMarking' components under Road View (category)
Test all props can be changed and the component update as expected.

~- [ ] Branch has been run in docker.~
- [x] Lint and ~test workflows~ pass.


src/RoadSurface/RoadSurface.types.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * The props needed to define a road surface
+ */
+export type RoadSurfaceProps = {
+  /**
+   * A list of numbers defining the coordinates of points along the outline of the road
+   * format: [x0, y0, z0, x1, y1, z1 ...]
```
arpadlomenipg, line 7: Do we need the z-coordinates? This should always be a 2D component

src/RoadMarking/RoadMarking.types.tsx
```diff
@@ -0,0 +1,23 @@
+export type RoadMarkingProps = {
+  /**
+   * The points representing the path of the marking [x0, y0, z0, x1, y1, z1, ...]
```
arpadlomenipg, line 3: Do we need the z-coordinates? This should always be a 2D component

src/RoadMarking/RoadMarking.tsx
```diff
@@ -0,0 +1,56 @@
...
+            dash={dash}
+            stroke={color}
+            points={pointsXY}
+          />
+          <Line
+            key={`${key}_cutout`}
+            strokeWidth={width * 0.5}
+            scaleY={-1}
+            stroke={color}
+            globalCompositeOperation="destination-out"
```
arpadlomenipg, line 36: I think this part requires some investigation. In storybook, it seems to me that the cutout is not transparent, but somewhere between the background color and the line's color. It can get difficult to notice e.g. in case of yellow line, see the posted image. If this is a problem, I would suggest to add another, optional prop "cutoutColor" (or something similar), which could define the color of the cutout.
yellowdoubleline

src/RoadSurface/RoadSurface.tsx
```diff
@@ -0,0 +1,23 @@
...
+  { points, color = "#78736e" }: RoadSurfaceProps,
+  key: React.Key
+) => {
+  return (
+    <Line
+      key={key}
+      scaleY={-1}
+      closed
+      fill={color}
+      points={points.filter(
```
arpadlomenipg, line 15: If z-coordinates will be removed, this should be rewritten

arpadlomenipg (changes requested):
Some changes requested. In general, we could probably just remove the z-coordinates, then all the storybooks should also be changed accordingly

src/RoadMarking/RoadMarking.tsx
AronPetrauskas: Seems like it is to do with the size of the component. I think it is only a display error/funny thing. I have tried both increasing the `width` of the line and increase the `scaling` of the stage and both of these methods resolve the weird shadow in the cutout. BTW yellow, is probably not the greatest color cause I can barely see the picture.

Anyways, at some size the display is proper. Thus the `global-composition` is working. As you can zoom in and out on the road viewer this should not be a huge issues. I think it is best to accept this behavior for now.
image

AronPetrauskas: For now I have increased the `scaling` on the stage, so the visual issue should not be visible on the default component

src/RoadSurface/RoadSurface.tsx
AronPetrauskas: Done

src/RoadMarking/RoadMarking.types.tsx
AronPetrauskas: Removed, only x and y now.

src/RoadMarking/RoadMarking.tsx
```diff
@@ -0,0 +1,53 @@
...
+            key={key}
+            strokeWidth={width * 2.5}
+            scaleY={-1}
+            dash={dash}
+            stroke={color}
+            points={points}
+          />
+          <Line
+            key={`${key}_cutout`}
+            strokeWidth={width * 0.5}
```
arpadlomenipg, line 30: The weird looking area between the lines is still there. I think it is because there is a transition from the end of the double lines to transparent, and if there is not enough space for the transition to happen, it gets blurry.

I would suggest to increase this value to 0.8 (barely visible) or 0.9, to avoid having this. I guess it depends whether that still looks acceptably close, I am just afraid of having an ugly blurred area between the lines.

arpadlomenipg (changes requested):

src/RoadMarking/RoadMarking.tsx
AronPetrauskas: I don't think this is the solution as it would only solve the blurred lines at a specific combination of `width` and `scale`. Since the end goal is to allow the user to zoom in or zoom out, the blurry border would be notable at some zoom.

Furthermore, over a dark background like a road the blurred bit is barely visible. Additionally, I have tried changing the `globalCompositeOperation` to `xor` this results in the same behavior. I have also tried to make the 'middle' line a white line on top, same issue persists. I just don't think it is an issue since the user can zoom.

arpadlomenipg: My main concern is that once this gets to RESULT, the user will probably zoom onto the vehicle, to see events etc. Then this would be easily noticeable. I am not sure it is best to just leave it as is, but I guess it is a question of whether this is an important enough issue or not.

Since we seem to be on a deadlock on this one, it might be best to ask a third opinion on this one.

AronPetrauskas: The point I have been trying to make is specifically when you zoom this issue won't happen. You're using browser zoom, when the user zooms in the roadviewer we change the scale of the internal components of the canvas, thus this won't be visible.

AronPetrauskas: https://github.com/IPG-Automotive-UK/react-ui/assets/67595415/93400112-c2d0-4fe2-84a3-a523a683201f

AronPetrauskas: @jegasriskantha as DRI for scene could you advise please?

arpadlomenipg: The discussion ended up with it being OK.

arpadlomenipg (approved):

PR #869, opened 2024-05-13
TD 2737 Create React UI Bullet Gauge component

syedsalehinipg:
Closes [TD-2737](https://sce.myjetbrains.com/youtrack/issue/TD-2737)
Contributes to [TD-2030](https://sce.myjetbrains.com/youtrack/issue/TD-2030)

## Changes

This PR introduces a new `BulletGauge` component. This component is a gauge chart that can be used to represent data in a compact, linear format. It's particularly useful for comparing the measure of a primary metric against one or more other metrics.

Features:

- The gauge is rendered using the `react-plotly.js` library.
- The color of the gauge bar changes based on the value: red for values less than 30, green for values greater than 70, and orange for values in between.
- The component accepts title, value, and suffix as props.

Other changes:

- `SurfacePlot` and `LinePlot`
  - Changed interface to type
  - Updated JSDoc comments format as they weren't showing in Docs on storybook

## UI/UX

Light mode:
image

Dark mode:
image

## Testing notes

_Use case for this component would be VIRTO.DATA > Search > Details_ 

- Visual testing should be performed to ensure the component looks as expected in both light and dark modes.

- Use storybook to change prop values of value and suffix and see changes reflected. 

- Check changes based on the value: red for values less than 30, green for values greater than 70, and orange for values in between.

As playwright test scripts have been removed recently, to run the playwright tests locally you can add the scripts back in:
```
 "test:playwright": "pnpm exec playwright install --with-deps && pnpm run storybook:build && pnpm run storybook:serve:start && pnpm exec playwright test && pnpm run storybook:serve:stop",
 "test:playwright:headed": "pnpm exec playwright install --with-deps && pnpm run storybook:build && pnpm run storybook:serve:start && pnpm exec playwright test --headed && pnpm run storybook:serve:stop",
```

Then run:
`pnpm run test:playwright`
    

~~- [ ] Branch has been run in docker.~~

### DIFF
--- a/src/BulletGauge/BulletGauge.spec.tsx
+++ b/src/BulletGauge/BulletGauge.spec.tsx
@@ -1,30 +1,97 @@
 import { expect, test } from "@playwright/test";
 
-/**
- * Test to check that the bullet gauge is rendered correctly.
- */
-describe("BulletGauge", () => {
-  test("should render the title", async ({ page }) => {
-    await page.goto(
-      "http://localhost:6006/?path=/story/plots-bulletgauge--default"
-    );
+import { getIframeContext } from "../utils/common";
 
-    await expect(page.getByText("Data Maturity")).toBeVisible();
-  });
+test("should render the title", async ({ page }) => {
+  await page.goto(
+    "http://localhost:6006/?path=/story/plots-bulletgauge--default"
+  );
 
-  test("should render the value", async ({ page }) => {
-    await page.goto(
-      "http://localhost:6006/?path=/story/plots-bulletgauge--default"
-    );
+  // Get the iframe context
+  const iframe = await getIframeContext(page);
 
-    await expect(page.getByText("30")).toBeVisible();
-  });
+  // Wait for elements are loaded
+  await iframe.waitForTimeout(5000);
 
-  test("should render the suffix", async ({ page }) => {
-    await page.goto(
-      "http://localhost:6006/?path=/story/plots-bulletgauge--default"
-    );
+  // Get the text content of the element with the text 'Data Maturity'
+  const titleText = await iframe.textContent("text=Data Maturity");
 
-    await expect(page.getByText("%")).toBeVisible();
-  });
+  // Check if the title is rendered
+  expect(titleText).toBeTruthy();
 });
+
+test("should render the value", async ({ page }) => {
+  await page.goto(
+    "http://localhost:6006/?path=/story/plots-bulletgauge--default"
+  );
+
+  // Get the iframe context
+  const iframe = await getIframeContext(page);
+
+  // Wait for elements are loaded
+  await iframe.waitForTimeout(5000);
+
+  // Get the text content of the element with the class 'numbers'
+  const numbersText = await iframe
+    .locator(
+      "div.plot-container.plotly svg.main-svg g.indicatorlayer g.trace g.numbers text"
+    )
+    .textContent();
+
+  // Check if the value is rendered
+  expect(numbersText).toContain("30");
+});
+
+test("should render the suffix", async ({ page }) => {
+  await page.goto(
+    "http://localhost:6006/?path=/story/plots-bulletgauge--default"
+  );
+
+  // Get the iframe context
+  const iframe = await getIframeContext(page);
+
+  // Wait for elements are loaded
+  await iframe.waitForTimeout(5000);
+
+  // Get the text content of the element with the text '%'
+  const suffixText = await iframe.textContent("text=%");
+
+  // Check if the suffix is rendered
+  expect(suffixText).toBeTruthy();
+});
+
+// Define the test cases with the value and expected color
+const testCases = [
+  { expectedColor: "rgb(211, 47, 47)", value: 29 }, // red
+  { expectedColor: "rgb(237, 108, 2)", value: 30 }, // orange
+  { expectedColor: "rgb(237, 108, 2)", value: 70 }, // orange
+  { expectedColor: "rgb(46, 125, 50)", value: 71 } // green
+];
+
+// iterate over test cases
+for (const testCase of testCases) {
+  test(`should render the gauge with correct colour for value ${testCase.value}`, async ({
+    page
+  }) => {
+    // Navigate to the specific story or dynamically set the value if possible
+    await page.goto(
+      `http://localhost:6006/?path=/story/plots-bulletgauge--default&args=value:${testCase.value}`
+    );
+
+    // Get the iframe context
+    const iframe = await getIframeContext(page);
+
+    // Wait for elements to load
+    await iframe.waitForTimeout(5000);
+
+    // Use a specific path to locate the 'value-bullet' rect in the SVG
+    const rectStyle = await iframe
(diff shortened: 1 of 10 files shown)


src/BulletGauge/BulletGauge.tsx
```diff
@@ -0,0 +1,73 @@
...
+          gauge: {
+            axis: {
+              range: [null, 100],
+              tickfont: {
+                color: theme.palette.mode === "light" ? "black" : "white",
+                size: 12
+              }
+            },
+            bar: {
+              color: value < 30 ? "#f44336" : value > 70 ? "#4caf50" : "#ff9800"
```
lukemojo, line 24: We should probably avoid hardcoding colours into components. Do we have colours in our palette that would work, if not could these be added? Traffic light colours would likely have re-use in lots of places.

lukemojo (changes requested):

src/BulletGauge/BulletGauge.tsx
syedsalehinipg: We could make use of the `theme.palette` colours like this

```
            bar: {
              color:
                value < 30
                  ? theme.palette.error.main
                  : value > 70
                    ? theme.palette.success.main
                    : theme.palette.warning.main
            },
```

lukemojo: Yep I think that would be better

lukemojo (approved):

src/BulletGauge/BulletGauge.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    await expect(page.getByText("30")).toBeVisible();
+  });
+
+  test("should render the suffix", async ({ page }) => {
+    await page.goto(
+      "http://localhost:6006/?path=/story/plots-bulletgauge--default"
+    );
+
+    await expect(page.getByText("%")).toBeVisible();
+  });
```
dmbarry86, line 29: You should add some tests for the colouring logic.

src/BulletGauge/BulletGauge.tsx
```diff
@@ -0,0 +1,78 @@
+import { BulletGaugeProps } from "./BulletGauge.types";
```
dmbarry86, line 1: I think you are missing the ClientOnly wrapper to support SSR. Refer to the SurfacePlot as an example.

https://github.com/IPG-Automotive-UK/react-ui/blob/main/src/SurfacePlot/SurfacePlotClientOnly.tsx#L1-L19

https://github.com/IPG-Automotive-UK/react-ui/blob/0cc95cce1e3324582cebe767e766b77ffbc9b1ce/src/SurfacePlot/index.ts#L4

dmbarry86 (changes requested):
* Missing tests
* Missing SSR wrapper
* Check % is always required and document in the prop if so or add new props if not

src/BulletGauge/BulletGauge.types.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * This type defines the props for the BulletGauge component.
+ */
+export type BulletGaugeProps = {
+  /**
+   * The value to be displayed on the gauge.
+   */
+  value: number;
```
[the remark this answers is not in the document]
syedsalehinipg: I think the gauge should be left open for other indicators and not just percentages. I am not sure as of now what these could be. But I would suggest we leave it flexible. 

- I can add a limit (100) 
- I can update the comment for value to mention the thresholds

src/BulletGauge/BulletGauge.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  { expectedColor: "rgb(237, 108, 2)", value: 30 }, // orange
+  { expectedColor: "rgb(237, 108, 2)", value: 70 }, // orange
+  { expectedColor: "rgb(46, 125, 50)", value: 71 } // green
+];
+
+// iterate over test cases
+for (const testCase of testCases) {
+  test(`should render the gauge with correct colour for value ${testCase.value}`, async ({
+    page
+  }) => {
```
dmbarry86, line 102: You should use test.each syntax here.

https://vitest.dev/api/#test-each

src/BulletGauge/BulletGauge.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { getIframeContext } from "../utils/common";
+
+test("should render the title", async ({ page }) => {
+  // Navigate to the page
+  await page.goto(
+    "http://localhost:6006/?path=/story/plots-bulletgauge--default"
+  );
+
+  // Get the iframe context
+  const iframe = await getIframeContext(page);
```
dmbarry86, line 12: Do we really need to have these tests in Playwright? Seems to me like React Testing Library would be suitable for these. Also less to migrate over to Cypress.

src/BulletGauge/BulletGauge.tsx
```diff
@@ -0,0 +1,81 @@
...
+import Plotly from "react-plotly.js";
+import React from "react";
+import { useTheme } from "@mui/material";
+
+const BulletGauge = ({ title, value, suffix }: BulletGaugeProps) => {
```
dmbarry86, line 6: Missing JSDoc comments for the component. https://gist.github.com/mattcorner/cf4bfb4cd967bb6fe13d2433010d516a#file-example-tsx-L37-L47

dmbarry86 (changes requested):
* Missing comments
* Question over the use of Playwright
* Opportunities for reuse of iframe util in other tests?

src/utils/common.ts
```diff
@@ -12,3 +13,15 @@ export function uniqueSortedArray(array: string[], sortOrder?: "asc" | "desc") {
...
+export async function getIframeContext(page: Page) {
+  const iframeElement = await page.waitForSelector(
+    'iframe[title="storybook-preview-iframe"]'
+  );
+  const iframe = await iframeElement.contentFrame();
+  if (!iframe) {
+    throw new Error("Unable to find or access the iframe content.");
+  }
+  return iframe;
+}
```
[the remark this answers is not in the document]
syedsalehinipg: Yes, there are a lot of tests for other components that are duplicating code for this check. This is why I decided to create the helper function here. But changing them all will have to be a separate PR.

src/BulletGauge/BulletGauge.spec.tsx
syedsalehinipg: There is case for that, but my thoughts based on the nature of the test on this specific component is that RTL is designed for unit and integration test. The test are not in a real browser. Here the primary focus is on testing the rendered look and feel. RTL is great for testing internal logic and simple user interactions, but it doesn't fully show the visual details that you can only see in a real browser. Playwright allows us to ensure that all visual aspects are as expected, closely simulating real user use case.

syedsalehinipg: Also for this component, Plotly.js is particularly challenging to test due to its heavy reliance on browser features.

dmbarry86 (approved):

PR #873, opened 2024-05-22
TD-2770 ImageUploader loading state

m4manjesh:
Closes [TD-2770](https://sce.myjetbrains.com/youtrack/issue/TD-2770/Add-loading-state-to-ImageUploader)

Added a new loading state to  `ImageUploader` component for better user experinece based on the feedback received [here](https://github.com/IPG-Automotive-UK/virto/pull/729#issuecomment-2122615753)

## Changes

- Added `isUploading` prop to control loading state
- Added new loading state UI
- Added new `SingleFileLoading` story 
- Increased file size limit from `1MB`  to `1000MB`
- Added tests 
- Added new 2 different custom error messages

## UI/UX
 @Sowbhagya-ipg 
image-loading

New error messages
Screenshot (460)
Screenshot (461)

## Testing notes

- Run `pnpm storybook` on this branch
- Go to `ImageUploader` and check you can switch values of new `isUploading` prop
- When its true you should see line progress loading state UI with "ImageUploading" text

### DIFF
--- a/src/ImageUploader/ImageUploader.tsx
+++ b/src/ImageUploader/ImageUploader.tsx
@@ -12,7 +12,7 @@ export default function ImageUploader({
   subText = "A default image will be used if no image is uploaded",
   dropzoneText = "Drag & Drop an image file here or browse",
   isUploading = false,
-  maxFileSize = 1000000000,
+  maxFileSize = 10000000,
   onAdd,
   onDelete,
   selectedFiles = [],
--- a/src/ImageUploader/ImageUploader.tsx
+++ b/src/ImageUploader/ImageUploader.tsx
@@ -1,4 +1,4 @@
-import { Box, Stack, Typography, alpha } from "@mui/material";
+import { Box, LinearProgress, Stack, Typography, alpha } from "@mui/material";
 
 import FileUploadIcon from "@mui/icons-material/FileUpload";
 import { ImageUploaderProps } from "./ImageUploader.types";
@@ -11,7 +11,8 @@ export default function ImageUploader({
   titleVariant,
   subText = "A default image will be used if no image is uploaded",
   dropzoneText = "Drag 'n' drop an image file here, or click to select",
-  maxFileSize = 1000000,
+  isUploading = false,
+  maxFileSize = 1000000000,
   onAdd,
   onDelete,
   selectedFiles = [],
@@ -50,7 +51,7 @@ export default function ImageUploader({
         titleVariant={titleVariant}
         required={required}
         subText={subText}
-        showDelete={selectedFiles.length > 0}
+        showDelete={!isUploading && selectedFiles.length > 0}
         onDelete={() => handleDelete(selectedFiles[0])}
       />
       <Box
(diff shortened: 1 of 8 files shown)


jegasriskantha:
1000mb (1Gb) seems huge for an image file, also by changing the default here we will be changing the default everywhere that uses this. Is this what we want @Sowbhagya-ipg ?

src/ImageUploader/ImageUploader.tsx
```diff
@@ -11,7 +11,8 @@ export default function ImageUploader({
   titleVariant,
   subText = "A default image will be used if no image is uploaded",
   dropzoneText = "Drag 'n' drop an image file here, or click to select",
-  maxFileSize = 1000000,
+  isUploading = false,
+  maxFileSize = 1000000000,
```
jegasriskantha, line 15: not sure about the default maxFileSize here

jegasriskantha (commented):
Looks good, just one comment on the default file size.

jegasriskantha:
The linked comment also mentions a better warning message for when the file limit is reached.  This doesn't seem to be addressed here, will that be addressed in another PR?
`File size exceeds the limit of {X}Mb . X being the limit set by us.`

src/ImageUploader/ImageUploader.tsx
m4manjesh: I came to a conclution to set default maxFileSize to 1GB after a chat with @mattcorner to be consistent in VIRTO apps

m4manjesh:
> The linked comment also mentions a better warning message for when the file limit is reached. This doesn't seem to be addressed here, will that be addressed in another PR? `File size exceeds the limit of {X}Mb . X being the limit set by us.`

No, i think i missed it and i should address it here

src/ImageUploader/ImageUploader.tsx
mattcorner: For reference, I said that this is what every other ImageUploader in VIRTO is set to and at least this would be consistent. I think we need to agree a better limit across the board.

Sowbhagya-ipg:
> 1000mb (1Gb) seems huge for an image file, also by changing the default here we will be changing the default everywhere that uses this. Is this what we want @Sowbhagya-ipg ?

No that seems so huge. I am fine with some reasonable file sizes around 5Mb - 10Mb. But keep the texts in MB Format not in bytes.

jegasriskantha:
> > The linked comment also mentions a better warning message for when the file limit is reached. This doesn't seem to be addressed here, will that be addressed in another PR? `File size exceeds the limit of {X}Mb . X being the limit set by us.`
> 
> No, i think i missed it and i should address it here

Okay, I'll mark this as request changes till this is addressed

jegasriskantha (changes requested):
update required to handle user friendly size limit error message.

src/ImageUploader/ImageUploader.tsx
mattcorner: Most global websites have a 10MB limit https://support.buffer.com/article/617-ideal-image-sizes-and-formats-for-your-posts. I can't see a use case for exceeding this, certainly as a RUI default.

m4manjesh: Thanks @mattcorner , i think we should go with 10 MB for now.

m4manjesh:
> update required to handle user friendly size limit error message.

Done, updated the PR descriptionand screenshots as well

Sowbhagya-ipg:
> > update required to handle user friendly size limit error message.
> 
> Done, updated the PR description and screenshots
looks good to me

jegasriskantha (approved):
LGTM

PR #876, opened 2024-05-29
TrafficLight type fix

AronPetrauskas:
## Changes

TrafficLight type is now a union. This fixes typing issues and the component has inbuilt protection from passing a type that is not recognized.

- ~[ ] Branch has been run in docker.~


src/TrafficSign/TrafficSign.types.ts
```diff
@@ -163,7 +95,8 @@ export type TrafficSignProps = {
     | "urbanAreaEnd"
     | "warningPedestrianCrossing"
     | "weightLimit"
-    | "widthLimit";
+    | "widthLimit"
+    | string;
```
syedsalehinipg, line 99: The `| string` at the end of your union type makes the type effectively just string, as it allows any string value. Is that what you want?

AronPetrauskas: Yes, this is the desired behavior. The component is still protected internally from hard erroring. The purpose of this is since the type is sometimes a name + value like speedLimit20, which is created dynamically from the incoming JSON by concatenating two strings. This just results in TS issues.

syedsalehinipg (approved):

PR #881, opened 2024-06-03
TD-2799 Uploader incorrect error messages

m4manjesh:
Closes [TD-2799](https://sce.myjetbrains.com/youtrack/issue/TD-2801/Fix-File-and-Image-Uplaoders-error-messages)

- [ ] ~Appropriate tests have been added.~

### DIFF
--- a/src/Uploader/useUploader.ts
+++ b/src/Uploader/useUploader.ts
@@ -77,11 +77,31 @@ export default function useUploader({
     const bytesToMb = maxFileSize / 1024 / 1024;
     const limitSize = `${bytesToMb.toFixed()} MB`;
 
+    // check if the uploader is for images
+    const isImageUploader = Object.keys(acceptedFiles ?? {}).some(key =>
+      key.includes("image/")
+    );
+
     // set error message based on error code
-    const errorMessage =
-      errorCode === "file-invalid-type"
-        ? "File type must be a .gif, .jpg, .jpeg, .png, .webp"
-        : `File size exceeds the limit of ${limitSize}`;
+    let errorMessage = "";
+    switch (errorCode) {
+      case "file-invalid-type":
+        errorMessage = isImageUploader
+          ? "File type must be a .gif, .jpg, .jpeg, .png, .webp"
+          : fileRejection[0].errors[0].message;
+        break;
+      case "file-too-large":
+        errorMessage = `File size exceeds the limit of ${limitSize}.`;
+        break;
+      case "too-many-files":
+        errorMessage = `You can only upload ${filesLimit} file(s).`;
+        break;
+      case "file-too-small":
+        errorMessage = `File is too small. Please upload a larger file.`;
+        break;
+      default:
+        errorMessage = "An error occurred while uploading the file.";
+    }
 
     // set error message
     setRejectionMessage(errorMessage);


src/Uploader/useUploader.ts
```diff
@@ -77,11 +77,31 @@ export default function useUploader({
     const bytesToMb = maxFileSize / 1024 / 1024;
     const limitSize = `${bytesToMb.toFixed()} MB`;
```
dmbarry86, line 78: This code can be moved in to the appropriate case as it's not needed for all statements.

src/Uploader/useUploader.ts
```diff
@@ -77,11 +77,31 @@ export default function useUploader({
...
     // set error message based on error code
-    const errorMessage =
-      errorCode === "file-invalid-type"
-        ? "File type must be a .gif, .jpg, .jpeg, .png, .webp"
-        : `File size exceeds the limit of ${limitSize}`;
+    let errorMessage = "";
+    switch (errorCode) {
+      case "file-invalid-type":
+        errorMessage = isImageUploader
+          ? "File type must be a .gif, .jpg, .jpeg, .png, .webp"
```
dmbarry86, line 90: Are these file extensions fixed by the component? Always feels a bit dangerous to be hardcoding things like this.

src/Uploader/useUploader.ts
```diff
@@ -77,11 +77,31 @@ export default function useUploader({
...
+          : fileRejection[0].errors[0].message;
+        break;
+      case "file-too-large":
+        errorMessage = `File size exceeds the limit of ${limitSize}.`;
+        break;
+      case "too-many-files":
+        errorMessage = `You can only upload ${filesLimit} file(s).`;
```
dmbarry86, line 97: If filesLimit is 1 then just say file. If it's greater than 1 then say files. Just looks cleaner.

src/Uploader/useUploader.ts
```diff
@@ -77,11 +77,31 @@ export default function useUploader({
...
+        break;
+      case "file-too-small":
+        errorMessage = `File is too small. Please upload a larger file.`;
```
dmbarry86, line 100: Curious, how is this error ever thrown? How is a small file defined? Do we even have props for that in the component?

src/Uploader/useUploader.ts
```diff
@@ -77,11 +77,31 @@ export default function useUploader({
...
+        break;
+      default:
+        errorMessage = "An error occurred while uploading the file.";
```
dmbarry86, line 103: Would it not be better to throw the default message from the fileRejection object? That way if they add some new cases in the future, at least we throw something probably more meaningful.

dmbarry86:
Copying @mattcorner for awareness in case it affects anything he has done in the forthcoming v8 release where file picker syntax has changed slightly.

dmbarry86 (changes requested):
Few small changes.

dmbarry86:
> * Appropriate tests have been added.

Why have no tests been added?

src/Uploader/useUploader.ts
m4manjesh: Yes, these extensions are fixed for image uploader. We dont have have a `acceptedFiles` prop for this, that's why i hardcoded this.

src/Uploader/useUploader.ts
m4manjesh: Yes it is very unlikely as we dont have a prop for this. Had a chat with @Sowbhagya-ipg abou this, we have added this error message to align with others, in future if we add a prop we can update this to `File is too small. Please upload a file larger than {x}Mb`. May be it needed to mention as a commnent for future reference.

src/Uploader/useUploader.ts
mattcorner: react-dropzone already gives you this error message based on the `acceptedFiles` property. If the built in error message is not sufficient, then I suggest you build it from the `acceptedFiles` property here as well instead hard coding ImageUploader logic into a hook that doesn't know anything about the type of uploader it is used in.

mattcorner:
> Copying @mattcorner for awareness in case it affects anything he has done in the forthcoming v8 release where file picker syntax has changed slightly.

All good. The underlying `useUploader` hook being modified here always used the correct syntax. It was just the `FileUploader` component which exposed the two syntaxes.

Sowbhagya-ipg:
Just to understand, Why the error states have different color border?

m4manjesh:
> Just to understand, Why the error states have different color border?

It is just an issue with the screeshots. I grabbed them while leaving hover state, updated with new ones.

dmbarry86:
Diff view seems to have lots of unrelated file changes on, making it harder for me to re-review what has changed. Looks like you incorrectly merged main into this branch rather than into the release branch and then release branch into this. Please resolve that before I re-review.

image

dmbarry86 (changes requested):
Fix branch status

m4manjesh:
> Fix branch status

Fixed

dmbarry86:
> Fixed

Kind of. My diff view now shows nothing if I use the option "Show changes since your last review". At least we don't have other files. I'll just have to manually re-review everything again.

dmbarry86:
I suggest moving this example file outside of the tests and into a separate test data file. I had to scroll for about 3 weeks in VSCode to actually get to the tests.

image

src/FileUploader/FileUploader.test.tsx
```diff
@@ -29,6 +30,25 @@ const multipleFiles = [
   }
 ] as FileWithData[];
 
+/**
```
dmbarry86, line 33: This code looks identical between FileUploader and ImageUploader. That should be ringing alarm bells in your head and you should be making a single common function they both can use.

src/Uploader/useUploader.ts
dmbarry86: Presumably we should also be showing the sise limit in the error message like we do for the file-too-large. I don't like the fact you've got code here tha tcan never be hit, it's effectively dead code. We also can't test it presumably. I suggest you remove it.

dmbarry86 (changes requested):
- [ ] Fix code duplication with test utility
- [ ] Remove dead code for file-too-small
- [ ] Move test data outside of test file

src/utils/testUtils/createDtWithFiles.ts
mattcorner, line 1: Test utilities should live outside of the /src/ folder otherwise they get included in the production build. I think a test/utils.ts file already exists that you could include this with, or add a test/utils/ folder.

dmbarry86 (approved):

PR #883, opened 2024-06-05
Enhancement - Breadcrumbs Component

lukemojo:
## Changes

Adds a Breadcrumbs component.

This is essentially a style component that renders the Mui Breadcrumbs component and standardises the styling of children for Virto. 

- Add a breadcrumbs component
- Add tests and a story

This does not control what components are being used as children and this component should be moved to the monorepo when possible and use useMatches and Remix Link internally. Tests do not assert about styling because changing the theme will break tests and using a different theme in app will make it a useless test.

## UI/UX

Screenshot 2024-06-05 at 08 49 26

Screenshot 2024-06-05 at 08 49 26

## Testing notes

Have a quick view in storybook. 

### DIFF
--- a/src/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/Breadcrumbs/Breadcrumbs.tsx
@@ -1,7 +1,5 @@
 import {
-  Link,
   Breadcrumbs as MuiBreadcrumbs,
-  Typography,
   breadcrumbsClasses
 } from "@mui/material";
 
@@ -12,16 +10,13 @@ import React from "react";
 /**
  * Breadcrumbs component
  */
-function Breadcrumbs({
-  breadcrumbs,
-  children,
-  component = Link
-}: BreadcrumbsProps) {
+function Breadcrumbs({ children, ...rest }: BreadcrumbsProps) {
   return (
     <MuiBreadcrumbs
+      {...rest}
       aria-label="breadcrumbs"
       separator={<NavigateNextIcon fontSize="small" />}
-      // style overrides to style children and breadcrumbs
+      // style overrides to style children
       sx={{
         [`& .${breadcrumbsClasses.li} *`]: {
           color: "text.secondary",
@@ -36,26 +31,7 @@ function Breadcrumbs({
         }
       }}
     >
-      {
-        // render children or breadcrumbs
-        children ||
-          breadcrumbs?.map(({ to, label }, i) =>
-            // render last breadcrumb as Typography
-            // render other breadcrumbs as Links or component prop
-            i === breadcrumbs.length - 1 ? (
-              <Typography key={label}>{label}</Typography>
-            ) : (
-              <Link
-                component={component}
-                key={label}
-                href={to}
-                underline="hover"
-              >
-                {label}
-              </Link>
-            )
-          )
-      }
+      {children}
     </MuiBreadcrumbs>
   );
 }
--- a/src/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/Breadcrumbs/Breadcrumbs.tsx
@@ -21,6 +21,7 @@ function Breadcrumbs({
     <MuiBreadcrumbs
       aria-label="breadcrumbs"
       separator={<NavigateNextIcon fontSize="small" />}
+      // style overrides to style children and breadcrumbs
       sx={{
         [`& .${breadcrumbsClasses.li} *`]: {
           color: "text.secondary",
--- a/src/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/Breadcrumbs/Breadcrumbs.tsx
@@ -1,4 +1,9 @@
-import { Link, Breadcrumbs as MuiBreadcrumbs } from "@mui/material";
+import {
+  Link,
+  Breadcrumbs as MuiBreadcrumbs,
+  Typography,
+  breadcrumbsClasses
+} from "@mui/material";
 
 import { BreadcrumbsProps } from "./Breadcrumbs.types";
 import NavigateNextIcon from "@mui/icons-material/NavigateNext";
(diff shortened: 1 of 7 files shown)


mattcorner:
Target for this PR should be an agreed release branch with @syedsalehinipg.

src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,39 @@
...
+          color: "text.secondary",
+          textDecoration: "none"
+        },
+        [`& .${breadcrumbsClasses.li}:last-child *`]: {
+          color: "text.primary",
+          textDecoration: "none"
+        },
+        [`& .${breadcrumbsClasses.li}:not(:last-child):hover *`]: {
+          textDecoration: "underline"
+        }
```
mattcorner, line 31: There is some specific styling in [TEST](https://github.com/IPG-Automotive-UK/virto/blob/805c56b66408e4a11037527a67c48110fd831135/apps/test/app/routes/tests.%24id/route.tsx#L103-L116) to ensure that breadcrumbs can never flow over multiple lines. I think this is useful for all apps to protect against long document names:

TEST:
image
Storybook:
image

mattcorner (changes requested):

src/Breadcrumbs/Breadcrumbs.tsx
lukemojo: I've updated this but changed it slightly, I found the example in scene. I noticed that you can still break the breadcrumbs as it was only constraining the last child. I've updated this to allow the constraint of every breadcrumb.

mattcorner: Nice, I don't think we currently have a VIRTO case where we have more than two breadcrumbs but nice attention to detail 👊

mattcorner:
I noticed that none of the mui props were showing up in the auto-generated docs. Seems to be something in broke in Storybook 8 so fixed that up here.

src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Breadcrumbs component
+ */
+function Breadcrumbs({ children, ...rest }: BreadcrumbsProps) {
+  return (
+    <MuiBreadcrumbs
+      {...rest}
+      aria-label="breadcrumbs"
+      separator={<NavigateNextIcon fontSize="small" />}
```
mattcorner, line 18: Storybook is unable to detect this as the default value in the generated docs.

image

Seems like the following syntax works:

```suggestion
function Breadcrumbs({
  children,
  separator = <NavigateNextIcon fontSize="small" />,
  ...rest
}: BreadcrumbsProps) {
  return (
    <MuiBreadcrumbs
      {...rest}
      aria-label="breadcrumbs"
      separator={separator}
```

image

mattcorner (changes requested):

mattcorner:
@Sowbhagya-ipg can you sign off these breadcrumbs from a UI perspective please? The plan is all VIRTO apps would use these for consistent UI.

Sowbhagya-ipg:
> @Sowbhagya-ipg can you sign off these breadcrumbs from a UI perspective please? The plan is all VIRTO apps would use these for consistent UI.

@lukemojo gave me a demo on this. The tooltip for truncated text can be added. The result looks fine for me.

src/Breadcrumbs/Breadcrumbs.types.ts
```diff
@@ -0,0 +1,9 @@
+import { BreadcrumbsProps as MuiBreadcrumbsProps } from "@mui/material";
+import { PropsWithChildren } from "react";
+
+/**
+ * Breadcrumbs component props
+ */
+export type BreadcrumbsProps = PropsWithChildren<
+  Omit<MuiBreadcrumbsProps, "sx">
```
mattcorner, line 8: If we're excluding `sx`, it feels like we should probably also exclude `style`, `className`, and `classes`.

src/Card/SummaryCard/SummaryCard.tsx
```diff
@@ -212,39 +176,27 @@ function SummaryCard({
...
-                </Typography>
-              </Tooltip>
-            </div>
-          }
-          subheader={
-            <Tooltip
-              title={subtitle}
-              disableHoverListener={!subTitleSizeOverflow}
-            >
+            <TruncatedTooltip tooltip={title}>
```
mattcorner, line 179: If I have an overflowing title as well as an action button, the button gets pushed off the card.

image

src/NoWrapTypography/NoWrapTypography.tsx
mattcorner, line 1: While we're here, can we make the box with the border the same width as the cut-off in the story. Its there to be visually helpful but it is just confusing at the moment.

image

src/TruncatedTooltip/TruncatedTooltip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+      sx={{
+        width: "220px"
+      }}
+    >
+      <TruncatedTooltip component={Typography} {...args} />
+    </Box>
+  );
+};
+
+// const LinkTemplate: StoryFn<TruncatedTooltipProps> = args => {};
```
mattcorner, line 36: Remove leftover comment

src/TruncatedTooltip/TruncatedTooltip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export const Default: StoryObj<typeof TruncatedTooltip> = {
+  args: {
+    children: "This is a long text that will be truncated",
+    href: "This/can/take/an/href/prop"
+  },
+
+  render: TypographyTemplate
+};
```
mattcorner, line 45: Is this not more correct, and removes the additional template?

```suggestion
export const Default: StoryObj<typeof TruncatedTooltip> = {
  args: {
    children: "This is a long text that will be truncated",
    component: Typography,
    href: "This/can/take/an/href/prop"
  },

  render: Template
};
```

src/TruncatedTooltip/TruncatedTooltip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+      with a two-inch brush. Look around, look at what we have. 
+      Beauty is everywhere, you only have to look to see it.`,
+    component: Typography,
+    multiline: 3
+  },
+
+  render: Template
+};
+
+export const WrappedSection = args => {
```
mattcorner, line 84: I'm not sure I get this story, or why it needs its own render method. At the very least it should be using `args` so that they appear in the storybook UI.

Other than the section component, how is this different from the multiline story above?

src/TruncatedTooltip/TruncatedTooltip.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+    )
+  },
+
+  render: Template
+};
+
+export const LinkUsingComponent: StoryObj<typeof TruncatedTooltip> = {
+  args: {
+    children: "This is a long text that will be truncated",
+    component: Typography,
```
mattcorner, line 62: Should component be something like "a"? It looks like Typography doing some clever switching under the hood, maybe lets be explicit for demo purposes.

mattcorner (changes requested):

src/Card/SummaryCard/SummaryCard.tsx
```diff
@@ -158,8 +153,8 @@ function SummaryCard({
       <Card sx={{ height, width }}>
         <CardHeader
           sx={{
-            "& .MuiCardHeader-content": {
-              width: "100%"
+            [` .${cardHeaderClasses.content}`]: {
+              overflowX: "hidden"
             },
```
mattcorner, line 158: Rather than forcing a pixel width, we can make use of the fact the underlying card is flex, and force it to not overflow horizontally. This leaves the title and subtitle components to use the full available width whether there is an action button present or not.

With the previous pixel width based approach, I ended up with unused white space if my text overflowed, and I didn't have a menu.

mattcorner (approved):
LGTM, I fixed up the SummaryCard a little bit. Check it over and make sure you're happy before merge.

PR #889, opened 2024-06-11
Enhancement - Autocomplete Readonly

lukemojo:
Add the ability to disable the clear button
Add the ability to make the field readOnly
Add styles to match the designs and disabled styles when field is readOnly

## Changes

This adds the ability to disable clearing the autocomplete and create a read only field.

Read only solves an issue where an uncontrolled field will not submit its value if it is disabled. The styles have been updated to reflect the designs (looking like a disabled field).

Disabling the clear button has been done to satisfy Zod discriminated union. Using a discriminated union with an empty value has some strange behaviour. This ensures that we can set a default that can be changed but never cleared.

- Add readOnly prop and add disabled styling
- Add disableClearable prop

## UI/UX

Screenshot 2024-06-11 at 17 33 23

## Testing notes

A story has been added for each update on Autocomplete.


dmbarry86:
As you are here, please can you fix the stories so that margin, size and variant are rendered as dropdowns with the appropriate options rather than free text fields.

image

dmbarry86:
Default story doesn't seem to show a value for readOnly. Presumably the default is false. Please check through all stories and make sure the new props that have been added are showing the correct values.

image

src/Autocomplete/Autocomplete.tsx
```diff
@@ -18,9 +18,11 @@ import { isKeyValueOption } from "../utils/common";
 
 export default function Autocomplete<
   Value extends KeyValueOption | string,
-  Multiple extends boolean | undefined
+  Multiple extends boolean | undefined,
+  DisableClearable extends boolean | undefined
 >({
   defaultValue,
+  disableClearable,
```
dmbarry86, line 25: should this not have a default, presumably false so we don't break any existing behaviour.

dmbarry86 (changes requested):
Looks good. Couple of small bits to tidy up with the stories and a missing default value.

src/Autocomplete/Autocomplete.tsx
lukemojo: Typescript won't allow me to do this. The DisableClearable is a type parameter with a constraint on it. The only thing it will allow you to assign is undefined. It's identical to the Multiple type and prop and produces the same error.

dmbarry86: I'm a little bit confused why we need to be using generics for our own wrapper component if I'm honest. I can remove them and then get the behaviour I think we should have but maybe I'm missing something here. Maybe worth a quick call with @mattcorner and the two of us as I think he may have first developed the type definitions for this component.

lukemojo:
> As you are here, please can you fix the stories so that margin, size and variant are rendered as dropdowns with the appropriate options rather than free text fields.
> 
> image



> As you are here, please can you fix the stories so that margin, size and variant are rendered as dropdowns with the appropriate options rather than free text fields.

It might be worth noting that setting defaults in the component seems to prevent storybook from pulling through the correct types for the doc descriptions. I had a brief look into it but don't have an answer why yet.

src/Autocomplete/Autocomplete.tsx
lukemojo: It looks like the generic affects the type of various different props, and using the AutocompleteProps type from MUI  requires the parameters. I'm getting a few errors pop up when I try to remove them.

From @mattcorner , "the generic affects some of the other props, so for instance if multiple is true, then onChange accepts an array, otherwise it accepts a scalar".

Still happy to talk further, but it feels like it should be separate to this PR?

dmbarry86: Yes we should use generic when we derive props from autocomplete and just pass them on as our own but that doesn't mean our own component needs to use generics? I have some code stashed locally which I think works so we can have a call when you are both free. I'm ok to handle it elsewhere, as long as the component still has the same defaults as the current component.

dmbarry86:
ReadOnly story doesn't seem to show values for all other props.

image

DisableClearable story gives me this error.
image


The other already existing stories seem to work fine and show the new options you added.

lukemojo:
> ReadOnly story doesn't seem to show values for all other props.
> 
> The other already existing stories seem to work fine and show the new options you added.

Oversight on my part, they are now fixed and I've checked the props on all the stories.

dmbarry86 (approved):

PR #895, opened 2024-06-13
TD-2774 Label overflow bug fix

dmbarry86:
Contributes to [TD-2774](https://sce.myjetbrains.com/youtrack/issue/TD-2774/Flashing-labels)

## Dependencies

N/A

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,17 +2,15 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Debug Tests",
       "type": "node",
       "request": "launch",
-      "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/react-scripts",
-      "args": ["test", "--runInBand", "--no-cache", "--watchAll=false"],
-      "cwd": "${workspaceRoot}",
-      "protocol": "inspector",
-      "console": "integratedTerminal",
-      "internalConsoleOptions": "neverOpen",
-      "env": { "CI": "true" },
-      "disableOptimisticBPs": true
+      "name": "Debug Current Test File",
+      "autoAttachChildProcesses": true,
+      "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+      "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
+      "args": ["run", "${relativeFile}"],
+      "smartStep": true,
+      "console": "integratedTerminal"
     },
     {
       // Requires the extension Debugger for Chrome: https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome
--- a/src/Card/DetailCard/DetailCard.tsx
+++ b/src/Card/DetailCard/DetailCard.tsx
@@ -1,14 +1,10 @@
-import { Box, Button, Popover, Stack, Typography } from "@mui/material";
-import {
-  DetailCardHeaderProps,
-  DetailCardLabelStackProps,
-  DetailCardProps
-} from "./DetailCard.types";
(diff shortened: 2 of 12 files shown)


.vscode/launch.json
dmbarry86, line 1: I originally started off writing RTL tests until I realised I needed a browser for the resize observer. I tried to debug tests and noticed that our debug config is way out of date and is still setup for jest. I've updated it to support vitest which now allows any test file to be debugged. Surprised nobody picked up on this until now. Does nobody debug tests when developing?

src/LabelSelector/LabelChip/LabelChip.tsx
```diff
@@ -16,6 +16,7 @@ export default function LabelChip({
   return (
     <Chip
       {...props}
+      className="label-chip"
```
dmbarry86, line 19: Not directly using this anywhere yet but also made sense to add class for the base label chip component as I was here.

src/Card/DetailCard/DetailCard.tsx
dmbarry86, line 1: All changes in this file are removing the existing code for labels and replacing with the new LabelChipGroup component.

dmbarry86:
@Sowbhagya-ipg This is a bug-fix for flashing label component when labels overflow. I've created a deploy preview in VIRTO so you can check it out. VIRTO.MODEL contains a model that has labels which previously triggered the issue. Please check summary and details card look as you expect.

dmbarry86:
Appears something from v8 release branch merge has broken a test. I'll have a look.

syedsalehinipg (approved):
LGTM.

- Playwright tests passed fine. 
- SummaryCard and DetailCard working fine.
- Looks good in VIRTO.MODEL too.

Sowbhagya-ipg:
@dmbarry86 Can we reduce the left margin of 16px to 8px?

Sowbhagya-ipg:
https://github.com/IPG-Automotive-UK/react-ui/assets/114938041/dd89a34c-4e50-40f7-b800-15f4d8099eb8

The details sections label acts clickable, But it is not. We need to remove the hover states as well here

Sowbhagya-ipg:
Can we do something about this? The label is taking the highest width here and again this is not proper as it is trimming the text. 
Screenshot 2024-06-14 at 12 28 57 pm

dmbarry86:
> When clicking on the labels in the table it results in page error.

This appears to be an existing issue on QA. I don't think it's related to anything done here. I'll have a look at it through when I actually integrate this component in VIRTO formally.

dmbarry86:
> Can we do something about this? The label is taking the highest width here and again this is not proper as it is trimming the text.

I have just copied the existing behaviour. Sure, we can change it though. Let's have a quick call later to discuss your ideas.

Sowbhagya-ipg:
May be this could be a separate issue to this: But the labels Selection filter behaves differently in MODEL and FLEET. 
In MODEL, I could multi-select labels but in FLEET I cannot.

Sowbhagya-ipg:
Does this fix is only for MODEL? Because looks like SCENE still has the issue @dmbarry86 
Screenshot 2024-06-14 at 1 29 31 pm

dmbarry86:
> Does this fix is only for MODEL? Because looks like SCENE still has the issue @dmbarry86

So far it has only bene applied to MODEL in the deploy preview you are testing. Same fix will rollout everywhere though once we are happy.

dmbarry86:
> May be this could be a separate issue to this: But the labels Selection filter behaves differently in MODEL and FLEET.
> In MODEL, I could multi-select labels but in FLEET I cannot.

That's definitely out of scope.

PR #899, opened 2024-06-28
Bug - TD-2849 Fix Breadcrumb and Numberfield visual defect

lukemojo:
Closes [TD-2849](https://sce.myjetbrains.com/youtrack/issue/TD-2849/RUI-BreadCrumbs-truncating-all-list-elements) & [TD-2850](https://sce.myjetbrains.com/youtrack/issue/VE-74/Style-and-spacing-issues-around-NumberFields-in-8.0.0-pre-release)

## Changes

Changes breadcrumbs so only the final item is truncated and wrapped in a tooltip. Restores the default "normal" margin to NumberFields

- Update component to only use the truncated tooltip on the final item
- Update styles to prevent all other items from shrinking
- Add "normal" prop as default to NumberField

## UI/UX
Screenshot 2024-06-28 at 11 37 17

## Testing notes

Check the story to ensure only one the last breadcrumb is truncated. Check that the NumberField has a margin by default in storybook.

### DIFF
--- a/src/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/Breadcrumbs/Breadcrumbs.tsx
@@ -18,6 +18,19 @@ function Breadcrumbs({
   separator = <NavigateNextIcon fontSize="small" />,
   ...rest
 }: BreadcrumbsProps) {
+  /**
+   *  Map children and wrap the final item in a truncated tooltip with separators
+   */
+  const mappedChildren = Children.map(children, (child, i) =>
+    child && isValidElement(child) ? (
+      Children.count(children) === i + 1 ? (
+        <TruncatedTooltip>{child}</TruncatedTooltip>
+      ) : (
+        child
+      )
+    ) : null
+  );
+
   return (
     <MuiBreadcrumbs
       aria-label="breadcrumbs"
@@ -36,6 +49,10 @@ function Breadcrumbs({
             textDecoration: "none"
             // Ensure children of block level elements are inline for ellipsis rendering
           },
+          "&:last-child": {
+            flexShrink: 1
+          },
+          flexShrink: 0,
           overflow: "hidden"
         },
         [`& .${breadcrumbsClasses.li}:last-child *`]: {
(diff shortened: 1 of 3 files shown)


src/Breadcrumbs/Breadcrumbs.tsx
AronPetrauskas, line 1: Now indeed only the last child is truncated. There is another issue however, `itemsAfterCollapse`, `itemsBeforeCollapse` and `maxItems` all seem to do nothing and do not work as described in the docs.

src/NumberField/NumberField.tsx
AronPetrauskas, line 1: Not sure this is a bug or not, but I can type in a number into this field and then the title of the `NumberField` returns down to a state it is in when `NumberField` is empty. Refer to screenshot.
image

AronPetrauskas (changes requested):
The original issues to be addressed have been fixed, however while testing I have found some other issues that should probably be addressed here.

src/Breadcrumbs/Breadcrumbs.tsx
lukemojo: This was a story issue. Args are now being spread correctly and these props should now be function as expected

src/NumberField/NumberField.tsx
lukemojo: This is a browser bug in Firefox. MuI seems to deal with it by always collapsing the labels on NumberFields in their examples.  Is this an option @Sowbhagya-ipg? The other way to get around this is by controlling the field and explicitly writing code about what should or shouldn't be allowed in the field. 

If we can collapse the labels I will do that here otherwise this should probably be a separate ticket and PR.

AronPetrauskas (approved):
`Breadcrumbs` work as expected. I am assuming the `NumberField` issue in Firefox will be addressed later. I am approving the PR.

src/NumberField/NumberField.tsx
dmbarry86: Similar issues already raised https://sce.myjetbrains.com/youtrack/issue/TD-2740/Number-field-doesnt-error-if-non-numeric-value-entered

PR #900, opened 2024-07-01
Fix TabPanel overflow behaviour

AronPetrauskas:
## Changes

The way TabPanel treats its children is fixed, so that a scrollbar appears if the children overflow the parent component.

~- [ ] Branch has been run in docker.~
~- [ ] Appropriate tests have been added.~


lukemojo (changes requested):
LGTM for now. Can you just remove the pre-release.md from the PR (I assume you branched from main). It's not a huge deal but that release branch doesn't really need any more clutter. 

For anyone else viewing this, discussions were had about the approach of this component and the usage of children. For now we have decided to do the minimum to unblock sprint work, but this will be placed on a list of components to revisit.

lukemojo (approved):

PR #884, opened 2024-06-05
VE-47 React UI component improvements

syedsalehinipg:
Closes [VE-47](https://sce.myjetbrains.com/youtrack/issue/VE-47)
Contributes to [VE-33](https://sce.myjetbrains.com/youtrack/issue/VE-33)

## Changes

This PR addresses UI issues mentioned in [#755](https://github.com/your-repo/issues/755)

### TreeViewList
- Removed the gap between each child item.

### TabPanel
- Made the content area scrollable.
- Fixed the tabs' position to the top.

### SurfacePlot
- Added new props: `showGrid`, `labelFontSize`, `exponentFormat`, and `minHeight`.

### LinePlot
- Moved lazy import outside of the export client-only function.
- Added new props: `showGrid`, `labelFontSize`, `exponentFormat`, and `minHeight`.
- Set `markers` and `showGrid` to true by default.
- Set `minHeight` to 400 by default.
- Changed marker color and size to match the existing use case.
- Adjusted margin to match the existing use case.
- Changed grid color to match the existing use case.

## UI/UX

Before:
image

Now:
image

Before:
image

Now:
image

Before:
image

Now:
image

Before:
image

Now:
image

### TreeViewList
- Verify that the gap between each child item is removed.
  - Inspect the spacing visually and ensure that there is no unintended margin or padding between items.

### TabPanel
- Check that when the window is smaller than the content area, a scroll bar appears which only scrolls the content while the tabs remain fixed.
  - Resize the browser window to ensure the scroll bar appears.
  - Ensure that the tabs remain fixed at the top and do not scroll with the content.
  - Verify that the scrolling behavior is smooth and does not impact the usability of the tabs.

### SurfacePlot | LinePlot
- Check that the new props (`showGrid`, `labelFontSize`, `exponentFormat`, and `minHeight`) reflect the relevant changes. Use Storybook for testing these props.
  - `showGrid`: Verify that the grid visibility can be toggled on and off.
  - `labelFontSize`: Change the label font size and confirm the updates reflect accurately on the plot.
  - `exponentFormat`: Check that different exponent formats are applied correctly.
  - `minHeight`: Ensure that the minimum height of the plot is respected and adjust as expected.

### General
- Verify that existing functionality not related to these changes remains unaffected.

## Author checklist before assigning a reviewer

- [x] Reviewed my own code-diff.
- [ ] Branch has been run in docker.
- [x] PR assigned to me or an appropriate delegate.
- [x] Relevant labels added to the PR.
[...]

PR #909, opened 2024-07-09
VE-84 plot styling fixes

dmbarry86:
Closes [VE-84](https://sce.myjetbrains.com/youtrack/issue/VE-84/Revert-breaking-Plot-changed-in-RUI)

## Changes

The [v7.9.0 release](https://github.com/IPG-Automotive-UK/react-ui/releases/tag/v7.9.0) introduced some unintended breaking changes to the line plot and surface plot components. This reverts some of them made [here](https://github.com/IPG-Automotive-UK/react-ui/pull/884/files#r1666769902) as well as fixes some other existing issues.

Following changes were made:
* Removed the minHeight prop. The plot components should fill the available space of their parent, as before. Parent height can be controlled if desired.
* Removed hardcoded line and marker colors and switched to using the theme.
* Reset line width back to 2.
* Removed the exponentFormat prop that was added. We always want to be using "E" for scientific notation.
* Added new story to both plot components to show how the axis look with scientified notation as this is an important usecase to consider.
* Removed the showTitle prop. It was weird to have both a title prop, which can be empty, as well as a showTitle boolean and need to use both in conjunction. Now we just have title prop and empty title dictates not wanting to show a title.
* Renamed markers prop to showMarkers for consistency with the other showXXX props.
* Fixed showGrid prop which was only showing/hiding the vertical grid. It now controls the grid in all axes.
* Removed labelFontSize prop. We want to use font size 12 everywhere.
* Reinstated the margins with a small tweak to look good inside a card as we have in VIRTO. I've also added a new story to showcase the usecase inside a card to aid us in the future.

Note there is still a known issue with the vertical axis label being too close to the y axis tick labels when they are showing scientific notation. Some time was invested here to try and find a way to address that but nothing came up so far. We won't hold up these fixes for that and will handle it separately as it's not a new issue in V8.

## Dependencies

N/A

## UI/UX

Please see the Storybook preview to test the LinePlot and SurfacePlot components.

## Testing notes

Please use the Storybook preview to test the LinePlot and SurfacePlot components.

I've also created a deploy preview in VIRTO [here](http://k8s-deploypr-apigatew-01484b5990-883837032.eu-west-2.elb.amazonaws.com/), which can be used to check the plots in VIRTO.RESULT now look correct again. Previously they were broken due to the styling issues. New addition compared to main in VIRTO is the grid lines are now on by default for both axes as agreed with Sowbhagya.

- [x] Appropriate tests have been added. (N/A)


mattcorner:
> * Removed the showTitle prop. It was weird to have both a title prop, which can be empty, as well as a showTitle boolean and need to use both in conjunction. Now we just have title prop and empty title dictates not wanting to show a title.

I've realised that the use case for this is that we might not want to show a title when the plot is rendering, but we do want to show the title in the full screen dialog. This is the behaviour we previously had in RESULT, but now we must either always show the title (which is duplicated in card) or never see the title in the full screen dialog.

mattcorner (changes requested):
Just the title/showTitle point to address

dmbarry86:
I've added a new fullscreenTitle prop to LinePlot and SurfacePlot component as we spoke about yesterday. If title prop is set it is shown in the plot in collapsed view and not fullscreen view. If fullscreenTitle prop is set it is shown only in the dialog title when in fullscreen mode. This is essentially the same behaviour we had previously with title and showTitle but now more explicit about what's really going on.

I've pushed a new prerelease to the VIRTO deploy preview so it can be tested in RESULT [here](http://k8s-deploypr-apigatew-01484b5990-883837032.eu-west-2.elb.amazonaws.com/result/results/plots?testId=668d4f55a14d6e663c2885a1&simulationId=668d4f55a14d6e663c28859e).

mattcorner (approved):

PR #911, opened 2024-07-09
Enhancement - Add href support to FileDetails download button

lukemojo:
Add href and target props to FileDetails
Add story
Add test
Update types

Closes [VE-91](https://sce.myjetbrains.com/youtrack/issue/VE-91/Add-href-support-to-the-FileDetail-component)
Contributes to [TD-2863](https://sce.myjetbrains.com/youtrack/issue/TD-2863/Handle-api-errors-when-downloading-roads)

## Changes

Adds href support to the FileDetails component.

This introduces
- a downloadLinkHref prop
- a downloadLinkTarget prop
- Replaces the button with an anchor when the href is provided

## Testing notes

Check the file detail story that the button renders as an anchor and passes through the href. Ensure the other stories function with no change.

### DIFF
--- a/src/CardContent/FileDetails/FileDetails.types.ts
+++ b/src/CardContent/FileDetails/FileDetails.types.ts
@@ -12,7 +12,11 @@ export type FileDetailsProps = {
   /**
    * The download href link.
    */
-  downloadHref?: string;
+  downloadLinkHref?: string;
+  /**
+   * The download href link target attribute.
+   */
+  downloadLinkTarget?: string;
   /**
    * The fileTitle of the card.
    */
@@ -36,8 +40,4 @@ export type FileDetailsProps = {
    * An optional initial search term
    */
   search?: string;
-  /**
-   * The download link target attribute.
-   */
-  target?: string;
 };
--- a/src/CardContent/FileDetails/FileDetails.types.ts
+++ b/src/CardContent/FileDetails/FileDetails.types.ts
@@ -10,6 +10,10 @@ export type FileDetailsProps = {
    */
   downloadButtonTextOnSearch?: string;
   /**
+   * The download href link.
+   */
+  downloadHref?: string;
+  /**
    * The fileTitle of the card.
    */
   fileTitle: string;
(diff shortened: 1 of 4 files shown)


src/CardContent/FileDetails/FileDetails.types.ts
```diff
@@ -9,6 +9,14 @@ export type FileDetailsProps = {
...
    */
   downloadButtonTextOnSearch?: string;
+  /**
+   * The download href link.
+   */
+  downloadLinkHref?: string;
+  /**
+   * The download href link target attribute.
+   */
+  downloadLinkTarget?: string;
```
mattcorner, line 19: Is there a use case for target? I understood that SCENE would open the download url and then redirect you back to the current page. When would we want that to be in a new tab?

Just checking we're not adding unnecessary support here.

src/CardContent/FileDetails/FileDetails.types.ts
```diff
@@ -9,6 +9,14 @@ export type FileDetailsProps = {
    * The download button text on search.
    */
   downloadButtonTextOnSearch?: string;
+  /**
+   * The download href link.
+   */
+  downloadLinkHref?: string;
```
mattcorner, line 15: I think a lesson for next time we write a component like this is that nested props are a good idea. e.g. downloadLinkProps: { text, href, target, onClick} etc.

Too late to change now.

src/CardContent/FileDetails/FileDetails.types.ts
mattcorner, line 19: I don't think string is the correct type here. It's actually a string union. Seems like react has a `HTMLAttributeAnchorTarget` type somewhere we should presumably be able to access.

image

mattcorner (changes requested):

src/CardContent/FileDetails/FileDetails.types.ts
lukemojo: Perhaps. I figured if we're giving href support and using an anchor then we should give basic attribute support. I don't think they will be using it for the current story.

mattcorner (approved):

PR #914, opened 2024-07-18
VE-69 July dependency updates

mattcorner:
Closes [VE-69](https://sce.myjetbrains.com/youtrack/issue/VE-69/React-UI-dependencies-July)

## Changes

* Updates all dependency versions except `eslint@9.x` which is currently incompatible with some other dependencies e.g. `@typescript-eslint/eslint-plugin`
* Modified the way in which we theme the docs container to accomodate dark/light mode. The old method was no longer compatible with latest storybook.
* Fixed a bunch of console.error and console.warning in tests that have been ignored for a long time.

## Dependencies
- Merging to main as requested by @lukemojo. There is nothing behaviour changing here that needs to go in a specific release.

## Testing notes

No behaviour change expected.

## Author checklist before assigning a reviewer

- [x] Reviewed my own code-diff.
- [x] Branch has been run in docker.
- [x] PR assigned to me or an appropriate delegate.
- [x] Relevant labels added to the PR.
- [x] ~Appropriate tests have been added.~
- [x] Lint and test workflows pass.

PR #915, opened 2024-07-18
add dark mode styling for preview and loading containers

mattcorner:
Resolves some feedback noted in #914 

## Changes

- Adds storybook preview iframe styling for dark mode. This caters for the preview loading ui as well as the docs loading ui.
- Also updates how docs decide whether to render dark or light mode to avoid a temporary fallback while the state is undefined.

Not included:
- I cannot (yet) find a way to target the initial load styling which renders the iframe in white. I don't think its worth investing more time in this right now.

## Testing notes

When clicking between stories/docs in dark mode you should no longer experience a flash of white UI.

- [x] ~Appropriate tests have been added.~


.storybook/preview.tsx
```diff
@@ -1,12 +1,13 @@
+import "./styles.css";
+
 import { DARK_MODE_EVENT_NAME } from "storybook-dark-mode";
 import { DocsContainer } from "@storybook/addon-docs";
 import React from "react";
 import { addons } from "@storybook/preview-api";
+import { store } from "storybook-dark-mode/src/Tool";
```
mattcorner, line 7: Reusing some of the internal implementation of storybook-dark-mode is not ideal as it is prone to break at any time. That said, they don't expose a way to read from localstorage to initialise state like they do internally. This seems like the best bad option to avoid flickering a fallback theme.

.storybook/preview.tsx
```diff
@@ -21,16 +22,18 @@ export const parameters = {
   },
   darkMode: {
     dark: { ...themes.dark },
-    light: { ...themes.light }
+    light: { ...themes.light },
+    stylePreview: true
```
mattcorner, line 26: This applies the dark | light classname to the preview iframe `html` tag which then applies the correct styles.

.storybook/styles.css
```diff
@@ -0,0 +1,70 @@
+.dark {
```
mattcorner, line 1: This css styles the error/loading/docs loading UIs in a dark theme that matches the rest of the storybook theme.

dmbarry86:
What's the status on this one chaps? Been open for a while.

lukemojo (changes requested):
I'm getting an error when running the branch. 

Screenshot 2024-07-30 at 09 42 38

I've tried reinstalling and binning off the node_modules and reinstalling from scratch, but I get the same error. I checked out main to make sure it wasn't a me error, I saw no issues.

lukemojo (approved):
LGTM. Flashing is gone and no errors

PR #920, opened 2024-08-08
TD-2862 FileUploader extension validation

mattcorner:
Contributes to [TD-2862](https://sce.myjetbrains.com/youtrack/issue/TD-2862/Validate-file-extensions-in-the-RUI-Uploader-component)

## Changes

- We no longer render any file rejection messages on hover. This is because the DataTransferItem object we have access to during this event does not contain enough information to correctly validate the file. Instead we wait for the drop event.
- We now validate files based on their extension. react-dropzone internally will allow files that pass the mime-type check through so even if we say `{ acceptedFiles: { "text/plain": [".rd5"] } }`, `.txt` were still being accepted.

## Testing notes

- Check that you do not get any errors on hovering with files. Validation should now only occur on drop.
- Check that for instance the `{ acceptedFiles: { "text/plain": [".rd5"] } }` case no longer allows `.txt` files to be accepted. There is a new automated test to cover this case as well.

### DIFF
--- a/src/FileUploader/FileUploader.test.tsx
+++ b/src/FileUploader/FileUploader.test.tsx
@@ -1,5 +1,5 @@
 import { MultipleFiles, SingleFile } from "./__data__/uploaderTestFiles";
-import { describe, expect, test } from "vitest";
+import { describe, expect, test, vi } from "vitest";
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 
 import FileUploader from "./FileUploader";
@@ -196,5 +196,34 @@ describe("FileUploader", () => {
     await waitFor(() =>
       expect(dropzoneElement).toHaveTextContent("File type must be .rd5.")
     );
+  });
+
+  test("accepts case insensitive extensions", async () => {
+    const onAdd = vi.fn();
+    const { getByTestId } = render(
+      <FileUploader acceptedFiles={{ "text/plain": [".rd5"] }} onAdd={onAdd} />
+    );
+
+    // create a valid file with an uppercase extension
+    const files = [new File(["ipg1"], "ipg1.RD5", { type: "text/plain" })];
+
+    // create a data transfer object with the files
+    const data = createDtWithFiles(files);
+
+    // get the dropzone element
+    const dropzoneElement = getByTestId("dropzone-root");
+
+    // trigger the drop event
+    fireEvent.drop(dropzoneElement, data);
+
+    // expect the onAdd callback to be called
+    await waitFor(() => {
+      expect(onAdd).toHaveBeenCalledOnce();
+      expect(onAdd).toHaveBeenCalledWith([
+        expect.objectContaining({
+          file: files[0]
+        })
+      ]);
+    });
   });
 });
--- a/src/FileUploader/FileUploader.test.tsx
+++ b/src/FileUploader/FileUploader.test.tsx
@@ -174,4 +174,27 @@ describe("FileUploader", () => {
       )
     );
   });
+
+  test("display error when file extension is not accepted", async () => {
+    const { getByTestId } = render(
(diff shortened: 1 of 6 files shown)


dmbarry86:
As you're here can you please do the feature team a favour and tag on this small update? https://sce.myjetbrains.com/youtrack/issue/TD-2973

mattcorner:
> As you're here can you please do the feature team a favour and tag on this small update? https://sce.myjetbrains.com/youtrack/issue/TD-2973

I think its a little more complicated than just including it in this PR, but I'll open a separate PR and release them together 👍

arpadlomenipg (changes requested):
one small change requested

src/Uploader/useUploader.ts
```diff
@@ -105,6 +103,22 @@ export default function useUploader({
     setRejectionMessage(errorMessage);
   };
 
+  // validator for file extension
+  const validator = (file: File | DataTransferItem) => {
```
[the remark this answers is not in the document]
mattcorner: I linked a react-dropzone bug ticket below this line. The validator does actually run on drag (rightly or wrongly), and in the case of a drag the file is of type DataTransferItem. Typing this correctly here forces us to correctly only handle the drop case in the if statement below.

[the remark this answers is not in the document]
mattcorner: In general we shouldn't type things just for our use case, we should type them for how they actually work in practice. Then typescript forces us to handle edge cases correctly.

arpadlomenipg (approved):
LGTM

src/Uploader/useUploader.ts
```diff
@@ -68,6 +68,9 @@ export default function useUploader({
     onAdd && onAdd(newSelection);
   };
 
+  // get the accepted file type extensions
+  const acceptedFileExtensions = Object.values(acceptedFiles ?? {}).flat();
```
arpadlomenipg, line 72: please apply a .toLowerCase() to avoid case sensitivity. E.g. right now, ".PNG" is not accepted in the image uploader of FLEET because of this

arpadlomenipg (changes requested):
I'm sorry to revert my previous approval, but I noticed meanwhile, that the `FileUploader` and `ImageUploader` is now case sensitive for extensions

src/Uploader/useUploader.ts
```diff
@@ -132,3 +147,22 @@ export default function useUploader({
...
+ *
+ * @example
+ * const fileName = "example.txt";
+ * const extension = getFileExtension(fileName);
+ * console.log(extension); // Output: ".txt"
+ */
+function getFileExtension(fileName: string) {
+  const parts = fileName.split(".");
+  if (parts.length > 1) {
+    return "." + parts.pop();
```
arpadlomenipg, line 165: please also add a .toLowerCase() here

src/FileUploader/FileUploader.test.tsx
```diff
@@ -174,4 +174,27 @@ describe("FileUploader", () => {
...
+
+    // trigger the drop event
+    fireEvent.drop(dropzoneElement, data);
+
+    // wait for the error message to be displayed
+    await waitFor(() =>
+      expect(dropzoneElement).toHaveTextContent("File type must be .rd5.")
+    );
+  });
 });
```
arpadlomenipg, line 200: could you please add a unit test for this file, where the `acceptedFiles` is having value ".rd5", but the file name is having ".RD5" extension, so we have a safeguard for case sensitivity in the future if we change something in the code?

Same in the `ImageUploader`, but with value ".png" and ".PNG"

arpadlomenipg (changes requested):
the issue is resolved, but I think it might be a good idea to add some unit tests for case sensitivity, in case we change the code again in the future

arpadlomenipg (approved):
LGTM

PR #921, opened 2024-08-09
TD-2973 Add disabled prop to FileUploader

mattcorner:
Closes [TD-2973](https://sce.myjetbrains.com/youtrack/issue/TD-2973/Update-RUI-FileUploader-to-add-disabled-prop)

## Changes

- Adds a `disabled` property to the FileUploader component. When true, this removes and deletion button, disables any hover effects on the dropzone, and disables any ability to drag/drop files.
- Also fixed an issue with the FileUploader in darkmode where text was always white, even when in an error state.

## UI/UX

Empty
image
image

Single file
image
image

Multi file
image
image

## Testing notes

- Test that disabled renders and behaves correctly for empty / single file / multiple file. The Figma can be found [here](https://www.figma.com/proto/0CVK945S1cs1ku0sUXcgv7/VIRTO.SCENE?node-id=2875-43915&t=MrVx5Q7gmwWRp35R-1). You can set the disabled flag using the story controls.

### DIFF
--- a/src/FileUploader/FileUploader.stories.tsx
+++ b/src/FileUploader/FileUploader.stories.tsx
@@ -52,6 +52,7 @@ export default meta;
 type Story = StoryObj<typeof FileUploader>;
 export const Default: Story = {
   args: {
+    disabled: false,
     dropzoneText: "Drag & Drop a file here or browse",
     filesLimit: 1,
     isValidating: false,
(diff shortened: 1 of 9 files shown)


mattcorner:
> LGTM. Did @Sowbhagya-ipg get a chance to review this?

We had a demo call earlier where we found one issue that is resolved. Could you give us a thumbs up if you are happy please @Sowbhagya-ipg? Thanks!

Sowbhagya-ipg:
Tested it. Looks fine @mattcorner 
Do we need a validation disabled status? While it is not necessarily a bad idea to keep it. But I think we won't end up with such status for validation.

mattcorner:
> Tested it. Looks fine @mattcorner Do we need a validation disabled status? While it is not necessarily a bad idea to keep it. But I think we won't end up with such status for validation.

Yeah, I don't think its a combination that a user could end up in. We only set the validating flag when a user drops a file, and they can't drop a file when its disabled. I'll leave it as is.

PR #923, opened 2024-08-12
TD 2959 Add new props and stories

m4manjesh:
Closes [TD-2959](https://sce.myjetbrains.com/youtrack/issue/TD-2959/RUI-updates)

## Changes

- `EditLabelDialog` moved to Dialog folder and exported to make it independent component.
- Added two new stories for default and edit.
- Whitespace trimming for name and description handled. Save button only enabled if name has valid values and if description adding and it also has a vlaid value. 
- `selected` and `description` props added for `LabelChip`
- Displaying tick icon if `selected` is true
- Showing Tooltip if  `description` passed
- Stories added for `selected` and `tooltip`
- Tests updated for new props

## UI/UX
@Sowbhagya-ipg 
Figma: https://www.figma.com/proto/0CVK945S1cs1ku0sUXcgv7/VIRTO.SCENE?node-id=9750-45959&t=HEeK6Oiv4MdDIhJA-0&scaling=min-zoom&content-scaling=fixed&page-id=2875%3A43915

EditLabelDialog
Screenshot 2024-08-12 132703
Screenshot 2024-08-12 132049

LabelChip with selected and tooltip
Screenshot 2024-08-12 132150
Screenshot 2024-08-12 132125

## Testing notes

- Check `EditLabelDialog` is visible under `Dialog` folder and works as expected 
- Check callbacks are fired and whitespace trimming works.
- Check `LabelChip` new props shows selected ico and tooltip as expected


src/EditLabelDialog/EditLabelDialog.jsx
mattcorner, line 1: image

Storybook docs are not being populated with any descriptions. As far as I know for JS components these are auto generated based the jsdoc descriptions in the propTypes at the bottom of this file.

src/EditLabelDialog/EditLabelDialog.jsx
```diff
@@ -24,7 +24,7 @@ export default function EditLabelDialog({
   onNew = () => {},
   onEdit = () => {},
   onClose = () => {},
-  labelDialogTitle = "Edit Label",
+  labelDialogTitle = "Add Label",
   label = { _id: "", color: "#005FA8", description: "", name: "" },
```
mattcorner, line 28: It's weird for a component to have a default label in this context. Shouldn't label be a **required** property?

src/EditLabelDialog/EditLabelDialog.stories.jsx
mattcorner, line 1: In the story controls, toggling `isOpen` has no effect:

image

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+// Default
+export const Default = {
+  args: {
+    label: {
+      _id: "",
+      color: "#005FA8",
+      description: "",
+      name: ""
+    }
```
mattcorner, line 50: You're missing some defaults here e.g. labelDialogTitle has a default in the component of "Add Label"

src/EditLabelDialog/EditLabelDialog.jsx
```diff
@@ -24,7 +24,7 @@ export default function EditLabelDialog({
   onNew = () => {},
   onEdit = () => {},
   onClose = () => {},
-  labelDialogTitle = "Edit Label",
+  labelDialogTitle = "Add Label",
```
mattcorner, line 27: A better property name would be `title`. The component name already contains Label and Dialog.

src/EditLabelDialog/EditLabelDialog.jsx
```diff
@@ -108,7 +112,6 @@ export default function EditLabelDialog({
     setName("");
     setDescription("");
     setColor("#005FA8");
-    onClose(event, "save");
```
mattcorner, line 111: Why the change in behaviour here?

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -0,0 +1,67 @@
...
+export default {
+  component: EditLabelDialog,
+  title: "Dialog/EditLabelDialog"
+};
+
+const Template = args => {
+  const [open, setOpen] = React.useState(false);
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
```
mattcorner, line 17: Rather than trying to handle the state yourself (and hence get out of sync with the storybook controls) we can use the `useArgs` hook from storybook to handle this state.

https://storybook.js.org/docs/addons/addons-api#useargs

See: https://github.com/IPG-Automotive-UK/react-ui/blob/5cb92d7a5eb62220004d8ee849d244560c179339/src/VehicleSelect/VehicleSelect.stories.tsx#L19-L31 for an example usage in RUI.

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const handleCancel = args => {
+    setOpen(false);
+    action("cancel")(args);
+  };
+  const handleSave = args => {
+    setOpen(false);
+    action("save")(args);
```
mattcorner, line 24: `action` is intended to reflect the function name that we are using. In this case `onClose`, `onEdit`, `onNew`. Currently you are showing `cancel` and `save`.

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -0,0 +1,67 @@
...
+export const EditLabel = {
+  args: {
+    label: {
+      _id: "1",
+      color: "#005FA8",
+      description: "This is a description",
+      name: "Label Name"
+    },
+    labelDialogTitle: "Edit Label"
+  },
```
mattcorner, line 65: Save some code duplication and build upon the default args.

```suggestion
  args: {
    ...Default.args,
    labelDialogTitle: "Edit Label"
  },
```

src/LabelSelector/LabelChip/LabelChip.stories.tsx
```diff
@@ -79,3 +79,21 @@ export const Small = {
...
+export const Selected = {
+  args: {
+    ...Default.args,
+    selected: true
+  },
+
+  render: Template
+};
+
+export const Tooltip = {
```
mattcorner, line 92: Missing code doc describing the intention of the story. What use case are we showing the developer with these stories.

src/LabelSelector/LabelChip/LabelChip.stories.tsx
```diff
@@ -79,3 +79,21 @@ export const Small = {
 
   render: Template
 };
+
+export const Selected = {
```
mattcorner, line 83: I expected this story to allow me to click on the LabelChip to select/deselect it. I think this is possible with the `onClick` property but it would be good to demonstrate this use case as per the Figma.

src/LabelSelector/LabelChip/LabelChip.tsx
```diff
@@ -12,31 +13,36 @@ export default function LabelChip({
...
-        [`& .${chipClasses.deleteIcon}:hover`]: {
-          color: theme => theme.palette.getContrastText(color)
-        },
-        visibility: visible ? "visible" : "hidden"
-      }}
-      label={<NoWrapTypography variant="inherit">{label}</NoWrapTypography>}
-      size={size}
-      variant={variant}
-    />
+    <Tooltip title={description}>
```
mattcorner, line 22: image
image

The Figma shows that the tooltips should be left aligned to the LabelChip (at least in the Label Selection popover), but they appear centered in Storybook. @Sowbhagya-ipg can you confirm the expectation here?

mattcorner, line 22: The font size for the tooltip in Figma is also different, although we are using the MUI default here which matches tooltips across the rest of our components. What do you want here @Sowbhagya-ipg?

src/EditLabelDialog/EditLabelDialog.jsx
mattcorner, line 27: Why the change in default here? Does this affect any other usages?

src/LabelSelector/LabelChip/LabelChip.test.tsx
```diff
@@ -77,4 +77,31 @@ describe("LabelChip", () => {
...
+    render(<LabelChip label="Label" selected />);
+
+    // get the selected icon
+    const selectedIcon = screen.getByTestId("DoneIcon");
+
+    // check that the selected icon is rendered
+    expect(selectedIcon).toBeInTheDocument();
+  });
+  // renders tooltip on hover of label chip
+  it("renders tooltip on hover", async () => {
```
mattcorner, line 91: One of the requirements is that no tooltip is rendered when there isn't a description. Can we add a test that captures this?

mattcorner (changes requested):

src/EditLabelDialog/EditLabelDialog.jsx
```diff
@@ -197,11 +200,12 @@ export default function EditLabelDialog({
...
           color="primary"
           disabled={
-            name.length === 0 ||
+            name.trim().length === 0 ||
             color.length === 0 ||
             !isLabelNameValid ||
             !hasChanged ||
-            !isLabelLengthValid
+            !isLabelLengthValid ||
+            !isDescriptionValid
```
mattcorner, line 208: Me and @Sowbhagya-ipg have some reservations about this logic. Currently I can have a name of "       name" and you pass this on in the `onSave` and `onEdit` callbacks. We should be trimming before returning these callbacks so that while the component shows "     name", the callback receives "name".

The save button should not be disabled when a user has entered a description of "             " as it is not a required field, and it will be trimmed down to an empty description in the callbacks which is valid.

src/LabelSelector/LabelChip/LabelChip.tsx
Sowbhagya-ipg: Left align it. @m4manjesh

Sowbhagya-ipg: I have used the MUI Default too. I wonder why it is not the same. The tooltip font size that I have is 10px.

src/EditLabelDialog/EditLabelDialog.jsx
m4manjesh: I think it should be

src/LabelSelector/LabelChip/LabelChip.tsx
mattcorner: Looks like we updated the default to be 14px 4+ years ago across all tooltips:

https://github.com/IPG-Automotive-UK/react-ui/blob/a742c9a1635c05fa7a631126dbfad76be1479be6/src/ThemeProvider/ThemeProvider.tsx#L69-L76

m4manjesh: Will wait @Sowbhagya-ipg to comment final decision on this.

src/LabelSelector/LabelChip/LabelChip.test.tsx
m4manjesh: added

src/LabelSelector/LabelChip/LabelChip.stories.tsx
m4manjesh: Story updated. Indroduced a new `showIcon` prop to decide to show the icon.

src/EditLabelDialog/EditLabelDialog.jsx
m4manjesh: Missed when editing, added it back

src/LabelSelector/LabelChip/LabelChip.tsx
Sowbhagya-ipg: The tooltip with a 10px is too small and 14px is relatively larger. I would suggest it to to be 12px which is lesser than the body font size we have and also readable. @m4manjesh @mattcorner and font weight

src/EditLabelDialog/EditLabelDialog.jsx
```diff
@@ -24,7 +24,7 @@ export default function EditLabelDialog({
   onNew = () => {},
   onEdit = () => {},
   onClose = () => {},
-  labelDialogTitle = "Add Label",
```
mattcorner: You need to update any uses of this:

e.g.https://github.com/IPG-Automotive-UK/react-ui/blob/336f7ab5cb9390c410f839a4d8031412f05df1bd/src/LabelSelector/LabelSelector.jsx#L279

image

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -4,36 +4,56 @@ import { Button } from "@mui/material";
...
 
 const Template = args => {
-  const [open, setOpen] = React.useState(false);
+  //  useArgs is a hook that returns the current state of the args object
+  const [{ isOpen }, updateArgs] = useArgs();
+
+  // update the args object with the new isOpen value
+  React.useEffect(() => {
+    updateArgs({ isOpen });
+  }, [isOpen, updateArgs]);
```
mattcorner, line 21: Do we need this? What case does this handle? I'm confused because isOpen comes from the useArgs hook already.

src/EditLabelDialog/EditLabelDialog.stories.jsx
```diff
@@ -0,0 +1,89 @@
...
+        onEdit={handleEditLabel}
+        onNew={handleAddLabel}
+      />
+    </>
+  );
+};
+
+// Default
+export const Default = {
+  args: {
```
mattcorner, line 64: You're still missing some default args. I'll list them for you:

```ts
{
options: [],
isOpen: false,
}
```

src/LabelSelector/LabelChip/LabelChip.tsx
```diff
@@ -12,31 +13,37 @@ export default function LabelChip({
...
-      label={<NoWrapTypography variant="inherit">{label}</NoWrapTypography>}
-      size={size}
-      variant={variant}
-    />
+    <Tooltip title={description} placement="bottom-start">
+      <Chip
+        {...props}
+        className="label-chip"
+        clickable={clickable}
+        icon={showIcon && selected ? <DoneIcon color="inherit" /> : undefined}
```
mattcorner, line 28: Whats the use case for selected without icon? This scenario wasn't accounted for in the original ticket so I'm wondering why we've added a third property.

mattcorner: Like now I can set `showIcon` to true, but no icon appears until I also set `selected` to true. Suggest we stick with the original `selected` proposal unless there is new information.

mattcorner (changes requested):

src/EditLabelDialog/EditLabelDialog.stories.jsx
m4manjesh: we dont need it

mattcorner (changes requested):

src/LabelSelector/LabelChip/LabelChip.test.tsx
```diff
@@ -77,4 +77,46 @@ describe("LabelChip", () => {
...
+  });
+  // test that the selected icon is not rendered when selected is false
+  it("does not render selected icon", () => {
+    render(<LabelChip label="Label" selected={false} />);
+
+    // there shouldn't be a selected icon
+    expect(screen.queryByTestId("DoneIcon")).not.toBeInTheDocument();
+  });
+  // renders tooltip on hover of label chip
+  it("renders tooltip on hover", async () => {
```
mattcorner, line 98: Your new tests are throwing a bunch of warnings;

https://github.com/IPG-Automotive-UK/react-ui/actions/runs/10370550853/job/28708693334#step:6:237

mattcorner (changes requested):

PR #927, opened 2024-08-22
TD-2989: Disabled prop to checkbox filter

Skoob1905:
Closes [TD-2989](https://sce.myjetbrains.com/youtrack/issue/TD-2989/Add-disabled-prop-for-the-checkbox-filter)

## Changes

- Add disabled prop to the checkboxFilter
- Added disabled to the default args so that this feature can be tested with both the alwaysOpen and normal usage states
- Also I reduced the options array for length 20 to 5, to avoid unnecessary scrolling.

## Dependencies

N/A

## UI/UX
This is the disabled state of the filter
image

## Testing notes

pull this pr locally, then run 'pnpm i; pnpm storybook', verify that this checkbox filter can be disabled.

- ~[x] Branch has been run in docker.~


src/Filter/CheckboxFilter/CheckboxFilter.test.tsx
AronPetrauskas, line 1: I am surprised that all the test pass, even though I can't select and option. Maybe the test checking that an option can be selected need to be rewritten.

src/Filter/CheckboxFilter/CheckboxFilter.tsx
```diff
@@ -73,31 +86,48 @@ function CheckboxFilterPopper({
...
   );
 }
 
 // renderer for a checkbox option
-function Option(props, option, { selected }) {
+function Option(
+  props: HTMLAttributes<HTMLLIElement>,
+  option: string,
+  { selected }: { selected: boolean },
+  { disabled }: Pick<CheckboxFilterProps, "disabled">
```
AronPetrauskas, line 113: I am not sure I understand why `selected` and `disabled` are wrapped in an object. I think either all are wrapped together, or none are wrapped.

src/Filter/AlwaysOpenAutocomplete.types.tsx
```diff
@@ -0,0 +1,8 @@
+import { AutocompleteProps } from "@mui/material";
+
+export type AlwaysOpenAutocompleteProps = AutocompleteProps<
+  string,
+  boolean,
+  boolean,
+  boolean
+>;
```
AronPetrauskas, line 8: Would it be possible to type this better, so we know what props are `string` and which are `boolean`?

AronPetrauskas (changes requested):
Currently, I can't select an option
https://github.com/user-attachments/assets/4be46a32-740c-48d8-9012-32b78b3a4c17

The `disable` toggle does work, but it seems that the original functionality of the component has been compromised

src/Filter/CheckboxFilter/CheckboxFilter.tsx
dmbarry86: They are not wrapped, they are destructured. The 3rd and 4th input args are objects. What we're doing here is just pulling out the selected and disabled fields from the objects in those 3rd and 4th arg positions.

src/Filter/CheckboxFilter/CheckboxFilter.test.tsx
dmbarry86: Probably means it's just a Storybook issue and not a component issue.

src/Filter/AlwaysOpenAutocomplete.types.tsx
dmbarry86: The typing is what it is here from the MUI component. We're just setting types for the four generics. We can add a comment against each to explain what it relates to. You can also just hover over the MUI type and it will explain what each arg is.

image

src/Filter/CheckboxFilter/CheckboxFilter.tsx
dmbarry86: You can refer to the MUI document for further explanation on the 4 args. https://mui.com/material-ui/api/autocomplete/

image

PR #928, opened 2024-08-27
TD-2998 Close label popper on click

dmbarry86:
Closes [TD-2998](https://sce.myjetbrains.com/youtrack/issue/TD-2998/LabelChipGroup-popover-stays-open-after-clicking-a-LabelChip)

## Changes

* Intercept the onClick callback for labels which are shown in the overflow popper inside LabelChipGroup so we can also close the popper when one of them is clicked.
* Updated the existing story for label chip group overflowing parent so labels are clickable to allow this to be tested.
* Updated the Playwright test to check popper is closed when we click on a label chip.

## Dependencies

N/A

## UI/UX

N/A

## Testing notes

Use Storybook to check the popper closes when you click a label. The overflowing parent story for label chip group can be used.

Run the Playwright tests. Note these need to be run locally as they are not currently part of the GitHub actions.

```
npx playwright install
npx playwright test LabelChipGroup
```

### DIFF
--- a/src/LabelSelector/LabelChipGroup/LabelChipGroup.tsx
+++ b/src/LabelSelector/LabelChipGroup/LabelChipGroup.tsx
@@ -2,6 +2,7 @@ import { Box, Button, Popover, Stack, Typography } from "@mui/material";
 
 import LabelChip from "../LabelChip/LabelChip";
 import { LabelChipGroupProps } from "./LabelChipGroup.types";
+import type { LabelChipProps } from "../LabelChip/LabelChip.types";
 import React from "react";
 import { ResizeObserver } from "@juggle/resize-observer";
 import { sortLabelChips } from "./sortLabelChips";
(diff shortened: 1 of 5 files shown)


src/LabelSelector/LabelChipGroup/LabelChipGroup.tsx
```diff
@@ -88,6 +89,20 @@ export default function LabelChipGroup({ chips }: LabelChipGroupProps) {
     }
   }, [chips, moreItemsRef, parentRef]);
 
+  // custom onClick event handler for label chips shown in the popper to close the popper on click
+  const handleOverflowingChipClick = (
+    event: React.MouseEvent<HTMLDivElement>,
+    onClick: LabelChipProps["onClick"]
```
arpadlomenipg, line 95: shouldn't we also check whether the LabelChip is clickable? Right now if I click an unclickable LabelChip, it will close the popover regardless. Or is there a reason we want it this way?

arpadlomenipg (changes requested):
I left opne comment about a doubt

src/LabelSelector/LabelChipGroup/LabelChipGroup.tsx
dmbarry86: How are you clicking an unclickable label?

arpadlomenipg: I set the `clickable` prop of the chips in storybook to false (tried it with just deleting the prop, works the same as expected), then when I click on the label, it still closes the LabelChipGroup, see the screen recording attached.

https://github.com/user-attachments/assets/cea08dd7-2f31-4f86-8a51-2867e244fb6d

dmbarry86: This sounds like an existing issue with the LabelChip component and/or the underlying MUI chip component. Seems strange that onClick is getting fired when clickable is set to false.

dmbarry86: Indeed it's actually the default behaviour of the MUI chip component it seems. https://mui.com/material-ui/api/chip/

image

I can update our usage of the MUI Chip component to change this behaviour.

arpadlomenipg: I was just texting the same out 😄 Sounds good, thanks 👍

dmbarry86: Seems like we also have an existing (badly written) test that is trying to test for this. It currently passes because the test is terrible.

image

arpadlomenipg (approved):
LGTM

PR #951, opened 2024-09-13
TD-3034 Remove redundant properties from RoadPreview types

jzhelyazkovtvm:
Closes/Contributes [TD-3034](https://sce.myjetbrains.com/youtrack/issue/TD-3034/Add-new-RoadPreview-component-to-RUI)

## Changes

Removed Label type from RoadPreview types and added the correct one, which is Label, but from LabelSelector component. Two of the properties in the removed type createdAt and updatedAt were redundant and causing problems in Virto, because this was not the appropriate type.

- Added the correct label type
- Removed createdAt and updatedAt from the stories as well

## Dependencies
N/A

## UI/UX
N/A

## Author checklist

Before I request a review:

- ~I have tested the changes in Docker / a deploy-preview.~
- ~I have included appropriate tests.~
- ~I have populated the deploy-preview with relevant test data.~

### DIFF
--- a/src/RoadPreview/RoadPreview.stories.tsx
+++ b/src/RoadPreview/RoadPreview.stories.tsx
@@ -54,10 +54,16 @@ export const Default = {
         name: "Test label 4"
       },
       {
-        _id: "66cde4855d45d85421f8620e",
-        color: "rgb(121, 0, 168)",
-        description: "desc for hi",
+        _id: "66cd24823d45d85421f8620c",
+        color: "#005FA8",
+        description: "desc for 1 hi",
         name: "Test label 3"
+      },
+      {
+        _id: "66cd24823f15d85421f8620s",
+        color: "#005FA8",
+        description: "last",
+        name: "last"
       }
     ],
     name: "SanFrancisco_AEB",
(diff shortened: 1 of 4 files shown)


AronPetrauskas:
In the author checklist, all items should be ticked before asking for a review. Here of course there are some that are not relevant. In this case you should cross those out ~like so~

AronPetrauskas (commented):
The bug fix looks good to me.

I looked through the tests for this component and saw that we are not testing this component with any Lables. I think we should have at least one test where we render the component with all props passed to it. Can @dmbarry86 advise if adding this extra test is necessary and if so should it be done as part of this pr?

jzhelyazkovtvm:
> The bug fix looks good to me.
> 
> I looked through the tests for this component and saw that we are not testing this component with any Lables. I think we should have at least one test where we render the component with all props passed to it. Can @dmbarry86 advise if adding this extra test is necessary and if so should it be done as part of this pr?

I didn't add any tests for labels, because it's a separated component which I am using and it has its own tests. However now I think that your idea is good, because of the following bug.

dmbarry86:
> I think we should have at least one test where we render the component with all props passed to it. Can @dmbarry86 advise if adding this extra test is necessary and if so should it be done as part of this pr?

I agree.

AronPetrauskas (approved):
LGTM!

PR #953, opened 2024-09-17
TD-3067 Add LazyLoadImage RUI component

arpadlomenipg:
Closes [TD-3067](https://sce.myjetbrains.com/youtrack/issue/TD-3067/)

## Changes

- Wraps the `Infographic` component in a lazy load structure to only load the images we need. The base of the solution is the `IntersectionObserver`, in this case the bounding box is the reference
- Adds a `LazyLoadImage` component, which can be used instead of `` tags to lazily load them. The base of the solution is the `IntersectionObserver`, observing the `` tag in question.

This feature can only be tested in the browser. However, the affected components `Infographic`, `DetailCard` and `SummaryCard` had tests using the react testing library, so those tests would fail (same would apply to the new component `LazyLoadImage`). After talking about this issue with @dmbarry86, we came to the conclusion that creating new playwright tests would be a wasted effort, as there is an ongoing migration of the playwright tests to cypress. So I just left TODO comments in the component code, and commented out the tests, and renamed it to `.txt` files to not have failed tests (an empty test file means a fail to vitest). For this same reason, there is also no test created for the new `LazyLoadImage` component.

## Testing notes

You can open the storybook from https://666c1309face6e7c27f20e04-kwrjebdwct.chromatic.com/

**From my experience, storybook is can have problems with this lazy loading, e.g. in Firefox it is not a lazy load for me, while in Chrome works as a lazy loading in storybook. Therefore, to test I recommend using the deploy-preview I created in Virto, see the details in this PR: https://github.com/IPG-Automotive-UK/virto/pull/1029**

In storybook, the following places are relevant: `LazyLoadImage`, `DetailCard`, `SummaryCard` and `Infographic`

In those places:
- open the Network tab in the browser's dev tools, filter for images
- open the story
- remember which images loaded
- scroll down to the bottom
- as you scroll down, you should notice in the network tab that new images are being loaded
- in the card components and Infographic: make sure if a chip is mentioned in the story, that it is indeed there, inside the image

## Author checklist

Before I request a review:

- ~~[ ] I have tested the changes in Docker / a deploy-preview.~~ -> not relevant for RUI
- ~~[ ] I have included appropriate tests.~~ -> see the reason in the Description section
- ~~[ ] I have populated the deploy-preview with relevant test data.~~ -> not relevant for RUI


src/CardContent/Infographic/Infographic.stories.tsx
```diff
@@ -27,7 +27,16 @@ export const Default = {
 
 export const WithVersionChip = {
   args: {
-    media: "https://picsum.photos/336/190",
+    media: "https://picsum.photos/336/191",
```
m4manjesh, line 30: I dont think we need a seperate story to test this, the 'default` is enough.

src/LazyLoadImage/LazyLoadImage.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (imgRef.current) {
+      observer.observe(imgRef.current);
+    }
+    return () => observer.disconnect();
+  }, []);
+
+  // render the image
+  return (
+    <img
+      alt={alt || ""}
```
m4manjesh, line 45: This works fine, Do you think adding a loader animation something like this https://mui.com/material-ui/react-skeleton/#pulsate-example for the image @Sowbhagya-ipg would be nice?

src/LazyLoadImage/LazyLoadImage.tsx
```diff
@@ -0,0 +1,53 @@
...
+      ref={imgRef}
+      loading="lazy"
+      src={isVisible ? src : undefined}
+      style={{ display: isVisible ? "block" : "none" }}
```
m4manjesh, line 49: i think this is not correct, we are passing style two time and something broken here. Shold be like this
```suggestion
      style={{
        ...ImgProps?.style,
        display: isVisible ? "block" : "none",
        width: "100%"
      }}
```

src/LazyLoadImage/LazyLoadImage.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+export default meta;
+
+const Template: StoryFn<LazyLoadImageProps> = args => {
+  return (
+    <>
+      <LazyLoadImage
+        {...{
+          ImgProps: {
+            style: { aspectRatio: "2/1", objectFit: "contain", width: "80%" }
```
m4manjesh, line 22: If i wrap the image in a div with fixed height and width with this style it going to look like below. 
Screenshot 2024-09-18 112335. We need the image fit to the outer div.

src/LazyLoadImage/LazyLoadImage.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+          },
+          alt: "alternative text 1",
+          src: "https://picsum.photos/336/191"
+        }}
+      />
+      <br />
+      <LazyLoadImage
+        {...{
```
m4manjesh, line 30: Add examples with image inside different size wrapper div

src/LazyLoadImage/LazyLoadImage.types.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export type LazyLoadImageProps = {
+  /**
+   * Source of the image
+   */
+  src: string;
+  /**
+   * Alternative text for the image
+   */
+  alt: string;
```
m4manjesh, line 14: Is `alt` really need to be mandatory @dmbarry86 ? Is so we dont need this check inside  component `alt={alt || ""}`

src/LazyLoadImage/LazyLoadImage.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+          alt: "alternative text 1",
+          src: "https://picsum.photos/336/196"
+        }}
+      />
+    </>
+  );
+};
+
+export const Default = {
+  render: Template
```
m4manjesh, line 93: add default to mandaroty props for docs

src/LazyLoadImage/LazyLoadImage.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+import LazyLoadImage from "./LazyLoadImage";
+import { LazyLoadImageProps } from "./LazyLoadImage.types";
+import React from "react";
+
+/**
+ * Wrapper to lazy load an image
+ */
+const meta: Meta<typeof LazyLoadImage> = {
+  component: LazyLoadImage,
+  title: "LazyLoadImage/LazyLoadImg"
```
m4manjesh, line 12: I think this should be under `General`

m4manjesh (changes requested):
Some question and suggestions, see commnets

src/LazyLoadImage/LazyLoadImage.tsx
arpadlomenipg: Generally I agree that the style should be spread, but I'm not so sure we want to set the width to 100%, I think that comes down to the usage of the component. This is a general lazy loader for any images, I don't think we should restrict the styling other than what is strictly necessary (which is the `display` prop)

src/LazyLoadImage/LazyLoadImage.stories.tsx
arpadlomenipg: I don't think this is something that we should handle inside the component. This is purely down to styling, if you were passing different styling props, then it would work the way you want to. I thought the idea here is that we use this component just like we would an ``, so the styling is as much a blank slate as possible, we expose the props, and those are freely changeable in VIRTO.

src/LazyLoadImage/LazyLoadImage.tsx
arpadlomenipg: @dmbarry86 any opinion about this?

src/LazyLoadImage/LazyLoadImage.stories.tsx
dmbarry86: I agree, this component is simple a wrapper for . This component shouldn't be making any decisions on styling, other than what it needs for the display hiding during load.

src/LazyLoadImage/LazyLoadImage.tsx
dmbarry86: I agree, this component is simple a wrapper for . This component shouldn't be making any decisions on styling, other than what it needs for the display hiding during load.

src/LazyLoadImage/LazyLoadImage.types.ts
dmbarry86: Yes, alt is mandatory for ``, otherwise you will get a warning. Whether that is implemented through mandatory type or optional type with some default, I don't mind.

src/CardContent/Infographic/Infographic.tsx
```diff
@@ -10,6 +10,10 @@ const Infographic = ({ media, version }: InfographicProps) => {
   // state to track if the image is visible
   const [isVisible, setIsVisible] = useState(false);
 
+  // states to track the status of the image load, these influence whether to display alt, or MUI Skeleton
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
```
arpadlomenipg, line 16: these were introduced to accomodate the `Skeleton`. If the element is visible (in viewport), but loading (not yet loaded and not yet errored), then we should show the Skeleton

src/CardContent/Infographic/Infographic.tsx
```diff
@@ -43,10 +50,13 @@ const Infographic = ({ media, version }: InfographicProps) => {
         width: 368
       }}
     >
+      {showSkeleton && <Skeleton width="100%" />}
       <CardMedia
         component="img"
-        image={isVisible ? media : undefined}
-        loading="lazy"
+        alt={hasError ? "card-infographic" : ""}
+        src={isVisible ? media : undefined}
```
arpadlomenipg, line 57: we incorrectly used the `image` prop until now, which is meant to be used if we are using `component="div"`. This had the side effect of not triggering the `onError` event. Fixed it by switching to src. I also added an `alt` text we missed until now

src/CardContent/Infographic/Infographic.tsx
```diff
@@ -43,10 +50,13 @@ const Infographic = ({ media, version }: InfographicProps) => {
...
+        onLoad={() => setIsLoaded(true)}
+        onError={() => setHasError(true)}
```
arpadlomenipg, line 59: Handlers in case an image finished loading or encounter an error, needed for the skeleton

src/LazyLoadImage/LazyLoadImage.tsx
```diff
@@ -14,11 +15,18 @@ import type { LazyLoadImageProps } from "./LazyLoadImage.types";
...
 }: LazyLoadImageProps) {
   // state to track if the image is visible
   const [isVisible, setIsVisible] = useState(false);
 
+  // states to track the status of the image load, these influence whether to display alt, or MUI Skeleton
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  // state variable to set the size for the skeleton to be displayed
+  const [dimensions, setDimensions] = useState({ height: 0, width: 0 });
```
arpadlomenipg, line 28: The logic in this file is the very same as in the above, so check those comments in doubt. This is the main difference, as we are observing here an image, we can get the size of the image from its reference, which I use to size the `Skeleton`. This had to be handled differently in the other file, as there we are observing the bounding box, which is a flex box.

src/LazyLoadImage/LazyLoadImage.tsx
```diff
@@ -34,20 +42,43 @@ export default function LazyLoadImage({
...
+          }}
+        />
+      )}
+      <img
+        alt={hasError ? alt : ""}
+        ref={imgRef}
+        src={isVisible ? src : undefined}
+        style={{
+          ...restStyle,
+          display: showSkeleton ? "none" : "block"
```
arpadlomenipg, line 76: another difference compared to how we are styling in `Infographic`. In Infographic don't need to worry about overwriting styling, as its styling is hard-coded.

Since this is a wrapper, styling comes from the props. I decontructed styling and use my own logic to make sure the display prop is doing what we need.

src/CardContent/Infographic/Infographic.tsx
```diff
@@ -1,35 +1,99 @@
...
       className="infographic-container"
       position={"relative"}
+      ref={boxRef}
       sx={{
         display: "flex",
         justifyContent: "center",
         width: 368
       }}
     >
+      {showSkeleton ? (
```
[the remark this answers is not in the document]
m4manjesh: This improves the loader UI/UX better

src/LazyLoadImage/LazyLoadImage.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const { style, ...restImgProps } = ImgProps || {};
+  const { display, ...restStyle } = style || {};
+
+  // render the image
+  return (
+    <>
+      {showSkeleton ? (
+        <Skeleton
+          sx={{
+            height: dimensions.height,
```
m4manjesh, line 72:
```
  borderRadius: 0,
  transform: "none"
```

may be similar style here as well

m4manjesh (changes requested):
See the comments

PR #955, opened 2024-09-25
TD-3098 Create StatusCountTable Component

jzhelyazkovtvm:
Closes/Contributes [TD-3098](https://sce.myjetbrains.com/youtrack/issue/TD-3098/Add-StatusCountTable-component)

## Changes

Created new RUI Status count table component, which will be used by Status Count Bar component for showing with delay information about the current statuses of a. The component is a table with title, and total statuses count in the header. Below as table rows are shown all statuses. Appropriate tests are added for the data, which populates the component and icons for the statuses.

- Created Status Count Table component

## Dependencies

N/A

## UI/UX

image
[Figma prototype ](https://www.figma.com/proto/WoHtIQKM6eLlOUuzK1zlpB/VIRTO.TEST?node-id=2853-11258&t=sfHgNt2O3kLh0HHA-1)

## Testing notes

N/A

## Author checklist

Before I request a review:

- ~[ ] I have tested the changes in Docker / a deploy-preview.~
- ~[ ] I have populated the deploy-preview with relevant test data.~

### DIFF
--- a/src/Status/StatusCountTable/StatusCountTable.tsx
+++ b/src/Status/StatusCountTable/StatusCountTable.tsx
@@ -56,7 +56,7 @@ export function StatusCountTable({ title, count }: StatusCountTableProps) {
                 key={row}
                 sx={{ "&:last-child td, &:last-child th": { border: 0 } }}
               >
-                <TableCell sx={{ pl: "16px" }} component="th" scope="row">
+                <TableCell sx={{ pl: 2 }} component="th" scope="row">
                   <Typography
                     variant="body2"
                     display="flex"
@@ -67,7 +67,7 @@ export function StatusCountTable({ title, count }: StatusCountTableProps) {
                     {statuses[row].label.text}
                   </Typography>
                 </TableCell>
-                <TableCell align="right" sx={{ pr: "16px" }}>
+                <TableCell align="right" sx={{ pr: 2 }}>
                   <Typography variant="body2">{count[row]}</Typography>
                 </TableCell>
               </TableRow>
--- a/src/Status/StatusCountTable/StatusCountTable.tsx
+++ b/src/Status/StatusCountTable/StatusCountTable.tsx
@@ -12,30 +12,19 @@ import React from "react";
 import { Status } from "../statuses.types";
 import { StatusCountTableProps } from "./StatusCountTable.types";
 import StatusIcon from "../StatusIcon";
+import statuses from "../statuses";
 
-// custom status count table component, which will be used in the status bar component on hover to show aditional data about statuses
-export function StatusCountTable({
-  title,
-  count,
-  tableSx = { borderRadius: "4px", boxShadow: 8, maxWidth: "280px" }
-}: StatusCountTableProps) {
+/**
+ * Table that renders statuses vs. the number of items that match that status. Header row shows the total count alongside a title.
+ */
+export function StatusCountTable({ title, count }: StatusCountTableProps) {
+  /** Calculate the total cound of the statuses */
   const totalCount = Object.values(count).reduce((a, b) => a + b, 0);
 
-  /** Render every status name with first letter Capital and replace dashes with empty space */
-  function capitalizeFirstLetterAndRemoveDashes(str: string): string {
-    // Handle empty string case
-    if (!str) return str;
-    // Capitalize the first letter of a status
-    const capitalized = str.charAt(0).toUpperCase() + str.slice(1);
-    // replace dashes with empty space
-    if (capitalized.includes("-")) {
-      return capitalized.replace("-", " ");
-    }
-    return capitalized;
-  }
-
   return (
-    <TableContainer sx={tableSx}>
+    <TableContainer
+      sx={{ borderRadius: "4px", boxShadow: 8, maxWidth: "280px" }}
+    >
       <Table size="small">
         <TableHead>
           <TableRow
@@ -44,12 +33,12 @@ export function StatusCountTable({
               height: "41px"
             }}
           >
-            <TableCell style={{ paddingLeft: "16px" }}>
+            <TableCell sx={{ pl: 2 }}>
               <Typography variant="subtitle2" data-testid="status-count-title">
                 {title}
               </Typography>
             </TableCell>
-            <TableCell align="right" style={{ paddingRight: "16px" }}>
+            <TableCell align="right" sx={{ pr: 2 }}>
               <Typography
                 variant="subtitle2"
                 data-testid="status-count-total-count"
@@ -67,11 +56,7 @@ export function StatusCountTable({
                 key={row}
                 sx={{ "&:last-child td, &:last-child th": { border: 0 } }}
               >
-                <TableCell
-                  style={{ paddingLeft: "16px" }}
-                  component="th"
-                  scope="row"
-                >
+                <TableCell sx={{ pl: "16px" }} component="th" scope="row">
                   <Typography
                     variant="body2"
                     display="flex"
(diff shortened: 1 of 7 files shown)


src/Status/StatusCountTable/StatusCountTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { StatusCountTableProps } from "./StatusCountTable.types";
+import StatusIcon from "../StatusIcon";
+
+// custom status count table component, which will be used in the status bar component on hover to show aditional data about statuses
+export function StatusCountTable({
+  title,
+  count,
+  tableSx = { borderRadius: "4px", boxShadow: 8, maxWidth: "280px" }
+}: StatusCountTableProps) {
+  const totalCount = Object.values(count).reduce((a, b) => a + b, 0);
```
mattcorner, line 22: Missing code doc

src/Status/StatusCountTable/StatusCountTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+    // Handle empty string case
+    if (!str) return str;
+    // Capitalize the first letter of a status
+    const capitalized = str.charAt(0).toUpperCase() + str.slice(1);
+    // replace dashes with empty space
+    if (capitalized.includes("-")) {
+      return capitalized.replace("-", " ");
+    }
+    return capitalized;
+  }
```
mattcorner, line 35: We shouldn't need to do any manipulation like this. The renderable string for a given status is listed in the config. You should read it from there. https://github.com/IPG-Automotive-UK/react-ui/blob/be9eb6826ecd86132598aea5b96ea49f6e5128aa/src/Status/statuses.ts

src/Status/StatusCountTable/StatusCountTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <TableContainer sx={tableSx}>
+      <Table size="small">
+        <TableHead>
+          <TableRow
+            sx={{
+              background: theme => theme.palette.background.default,
+              height: "41px"
+            }}
+          >
+            <TableCell style={{ paddingLeft: "16px" }}>
```
mattcorner, line 47:
```suggestion
            <TableCell sx={{ pl: 2 }}>
```

src/Status/StatusCountTable/StatusCountTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  component="th"
+                  scope="row"
+                >
+                  <Typography
+                    variant="body2"
+                    display="flex"
+                    alignItems="center"
+                    gap="8px"
+                  >
+                    <StatusIcon status={row as Status} width={24} height={24} />
```
mattcorner, line 81: We shouldn't need to coerce this type. The keys of the count object should be typed as Status instead of string.

This Status type lives here: https://github.com/IPG-Automotive-UK/react-ui/blob/58ff17b951d3edc0f64d953e923f28851df15775/src/Status/statuses.types.ts#L6

src/Status/StatusCountTable/StatusCountTable.tsx
mattcorner: Prefer using the sx prop and the mui shorthand. Applies to several places in this file.

src/Status/StatusCountTable/StatusCountTable.types.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type StatusCountTableProps = {
+  /**
+   * Title of the table
+   */
+  title: string;
+  /**
+   * Count is an object which contains objects with key status name and status count
+   */
+  count: { [status: string]: number };
```
mattcorner, line 11:
```suggestion
  count: { [status: Status]: number };
```

src/Status/StatusCountTable/StatusCountTable.types.ts
```diff
@@ -0,0 +1,16 @@
...
+  /**
+   * Optional style properties for the table element
+   */
+  tableSx?: SxProps<Theme>;
```
mattcorner, line 15: What is the use case for this prop? We didn't propose it in the planning.

src/Status/StatusCountTable/StatusCountTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+  TableRow,
+  Typography
+} from "@mui/material";
+
+import React from "react";
+import { Status } from "../statuses.types";
+import { StatusCountTableProps } from "./StatusCountTable.types";
+import StatusIcon from "../StatusIcon";
+
+// custom status count table component, which will be used in the status bar component on hover to show aditional data about statuses
```
mattcorner, line 16: We should use jsdoc syntax here as it then gets picked up by the storybook docs.

```suggestion
/**
 * Table that renders statuses vs. the number of items that match that status. Header row shows the total count alongside a title.
 */
```

mattcorner (changes requested):

mattcorner:
Don't forget to tag @Sowbhagya-ipg in RUI PRs and other PRs that involve UI so that she can give approval on the UI/UX aspects.

src/Status/StatusCountTable/StatusCountTable.types.ts
jzhelyazkovtvm: I was suggesting that we may need to reuse this component on another place and change the style.

src/Status/StatusCountTable/StatusCountTable.types.ts
jzhelyazkovtvm: I have received errors from typescript: An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead. I tried as well { [status in Status]: number } but then there is an error that status is not used.

Sowbhagya-ipg:
Will the abort be part of the count? I am asking this because I can't see it in the count list. @jzhelyazkovtvm

src/Status/StatusCountTable/StatusCountTable.types.ts
mattcorner: We prefer not to expose sx from components as they potentially hide real use cases that we should support. If we don't have a concrete use case now then lets remove for now.

src/Status/StatusCountTable/StatusCountTable.types.ts
mattcorner: Interesting, `Record<Status, number>` should work instead.

Sowbhagya-ipg:
The list order is important to this status component as we will be displaying the same order in reverse in the progress bar. @jzhelyazkovtvm Can you bring the same order of what we have in the Figma please?

mattcorner:
> Also I am not sure if we really need to count the sanity check status here as part of this status count table? @jzhelyazkovtvm @mattcorner Is this intentional?

This is showing that it supports all statuses that we have in RUI. In TEST we will only call this component with the statuses that TEST knows about which delivers your Figma. It is intentional to support all statuses because you hinted that we might use this view in FLEET for sanity checks too and we'd get that for free.

mattcorner:
> The list order is important to this status component as we will be displaying the same order in reverse in the progress bar. @jzhelyazkovtvm Can you bring the same order of what we have in the Figma please?

TEST will control the order

jzhelyazkovtvm:
> Will the abort be part of the count? I am asking this because I can't see it in the count list. @jzhelyazkovtvm

Yes, aborted will replace cancelled, this change was already implemented, but you can't see it, because the change is in another release. When merged it will be replaced

src/Status/StatusCountTable/StatusCountTable.tsx
jzhelyazkovtvm: After changing count property to Record<Status, number> we again need to coerce the type of icon text, as it doesn't recognizes it as a property of Type Status.

src/Status/StatusCountTable/StatusCountTable.types.ts
jzhelyazkovtvm: we can implement also as {[status in Status]: number }, but eslint throws an error that status in defined but never used

Sowbhagya-ipg:
UI looks good. @jzhelyazkovtvm

mattcorner (approved):

PR #969, opened 2024-10-22
VE-151 Added mutliselect option for variants

nikitharani:
Contributes [VE-151](https://sce.myjetbrains.com/youtrack/issue/VE-151/Add-Gateway-timing-functionality)

## Changes

- Update the vehicleSelector component logic to support multi-select for variants if multipleVariant is true.
- renamed multipleSelection to multipleGates
- Added limit tags prop
- Added story for multi-select variant
- Added tests for vehicleSelector component
- Added renderTags  with customizable label truncation and tag limits to AutoComplete Component
- removed pointerEvents it restored the default behavior similar to the original MuiAutocomplete component.

## UI/UX
@Sowbhagya-ipg  Please review the UI changes

image

Example of VehicleSelector Component integrated in VIRTO.VEHICLE remix 

vehicleSelector

## Testing notes

- User should able to select multiple Variants

## Author checklist

Before I request a review:

- [ ] I have tested the changes in Docker / a deploy-preview.
- [ ] I have populated the deploy-preview with relevant test data.
- [x] I have tagged a UI/UX designer for review if this PR includes UI/UX changes.

### DIFF
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -90,7 +90,6 @@ export default function Autocomplete<
       getOptionLabel={option =>
         isKeyValueOption(option) ? option.value : String(option)
       }
-      sx={{ pointerEvents: readOnly ? "none" : "auto" }}
       renderTags={
         limitTags
           ? (value, getTagProps) =>
@@ -108,10 +107,7 @@ export default function Autocomplete<
             className: readOnly ? "Mui-disabled label.Mui-disabled" : undefined
           }}
           InputLabelProps={{
-            ...params.InputLabelProps,
-            sx: {
-              pointerEvents: readOnly ? "none" : "auto"
-            }
+            ...params.InputLabelProps
           }}
           name={name}
           margin={margin}
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -91,7 +91,7 @@ export default function Autocomplete<
         isKeyValueOption(option) ? option.value : String(option)
       }
       renderTags={
-        limitTags
+        limitTags && limitTags > 0
           ? (value, getTagProps) =>
               limitTags && Tags(value, getTagProps, limitTags)
           : undefined
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -57,7 +57,7 @@ export default function Autocomplete<
         {value.slice(0, limitTags).map((option, index) => (
           <Chip
             {...getTagProps({ index })}
-            key={index}
+            key={isKeyValueOption(option) ? option.key : option}
             label={
               // truncate the tag label
               isKeyValueOption(option)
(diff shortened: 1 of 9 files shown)


src/Autocomplete/Autocomplete.tsx
```diff
@@ -43,6 +45,36 @@ export default function Autocomplete<
...
+    getTagProps: AutocompleteRenderGetTagProps,
+    limitTags: number
+  ) {
+    const numTags = value.length;
+    return (
+      <>
+        {value.slice(0, limitTags).map((option, index) => (
+          <Chip
+            {...getTagProps({ index })}
+            key={index}
```
mattcorner, line 60: Can we pick a better key than index?

src/VehicleSelect/VehicleSelect.tsx
```diff
@@ -11,6 +11,7 @@ function uniqueSortedArray(values: string[]) {
 // component to select a vehicle
 function VehicleSelect({
   flexDirection = "column",
+  limitTags = 1,
```
mattcorner, line 14: Did you mean to update `VehicleSelect`? It seems from the PR description that you were working on the `VehicleSelector` component?

src/VehicleSelector/VehicleSelector.types.ts
```diff
@@ -39,7 +39,7 @@ export type VehicleSelectorProps = {
   /**
    * Allow multiple selections
    */
-  multipleSelection?: boolean;
```
mattcorner, line 42: This is a breaking change. Did you find a specific case where you needed to support multipleGates = true and multipleVariants = false?

mattcorner: Ideally we would avoid a breaking change as you'll need to go and update everywhere that this is used, and target v9 instead.

If we do need to support variable combinations of `multiple` then I'd prefer that we did it in a nested object:

```ts
type Props = {
  /**
   * Allow multiple selections
   */
  multiple?: {
     /**
      * For the variant field
      */
     variant?: boolean;
     /**
      * For the gate field (if present)
      */
     gate?: boolean;
  }
}
```

mattcorner (changes requested):

src/VehicleSelect/VehicleSelect.tsx
nikitharani: Some tests were failing  for VehicleSelect without limitTag fixed that issue and removed changes from VehicleSelect component

src/VehicleSelector/VehicleSelector.types.ts
nikitharani: After discussing with Dave and Sowbhagya, decided to have only one prop multipleSelection for now.

src/Autocomplete/Autocomplete.tsx
```diff
@@ -58,7 +90,12 @@ export default function Autocomplete<
       getOptionLabel={option =>
         isKeyValueOption(option) ? option.value : String(option)
       }
-      sx={{ pointerEvents: readOnly ? "none" : "auto" }}
+      renderTags={
+        limitTags
```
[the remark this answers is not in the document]
mattcorner: Definitely it doesn't do the right thing for -1. Here I have 3 items selected, but it says I have 6.

image

mattcorner (changes requested):

PR #999, opened 2024-12-04
TD-3265 Do not ship sourcecode (.ts, .tsx)

mattcorner:
Contributes to [TD-3265](https://sce.myjetbrains.com/youtrack/issue/TD-3265/Plotly-incorrect-font)

## Changes

Until now we have shipped source code (.ts and .tsx) files as part of the build. We can't really remember why we did this because we also ship js code and .d.ts type decleration files generated by `tsc` which is all that should be needed. The result of shipping this source code is that running type-checks in consuming applications that set `skipLibCheck=true` also type checks raw typescript files in node_modules (but correctly skips .d.ts). This led to a number of false positives in the type-checking e.g. VIRTO.ID failing type-checks due to Plotly components that it does not use. This was fixed for example by installing type dependencies that the application itself did not need (but RUI did).

This PR removes the .ts and .tsx files from the build output. When integrated into VIRTO (see https://github.com/IPG-Automotive-UK/virto/pull/1219) we maintain application functionality, but are able to remove these extra unneeded dependencies and still pass type-checks.

## Testing notes

- Check that you can run `pnpm build`.
- Check that type-checks pass on all VIRTO apps that use RUI and the v9.0.4-0 pre-release. You can see this existing VIRTO PR https://github.com/IPG-Automotive-UK/virto/pull/1219 where this has been done.

## Author checklist

Before I request a review:

- ~I have populated the deploy-preview with relevant test data.~
- ~I have tagged a UI/UX designer for review if this PR includes UI/UX changes.~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,8 @@ jobs:
           cache: "pnpm"
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
+      - name: Test build
+        run: pnpm run build
       - name: Run tests
         run: pnpm run test:docker
       - name: "Report coverage"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "scripts": {
     "build": "pnpm run clean && pnpm run copy-files && tsc --project ./tsconfig.production.json",
     "clean": "rimraf dist",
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           cache: "pnpm"
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
+      - name: Test build
+        run: pnpm run build
```
mattcorner, line 33: Updated the test workflow to run a build to capture any issues before we release.

dmbarry86 (approved):

PR #1015, opened 2024-12-27
TD-3284 NoWrapTypography truncation ellipsis color issue

ShushN:
Closes/Contributes [TD-3284](https://sce.myjetbrains.com/youtrack/issue/TD-3284/NoWrapTypography-truncation-ellipsis-doesnt-match-color-of-child)

## Changes

- Ellipsis color matches the Link text (blue in both modes).
- Verified the fix in both light and dark modes.
- Replaced NoWrapTypography with TruncatedTooltip

Maybe a short description of what is not included and will come in future work where appropriate.

## Dependencies

NA

## UI/UX

Before in dark mode the ellipsis is white and in light mode it is black
Screenshot (3)
After in both mode ellipsis takes the same shade of blue as the link text
Screenshot (4)

## Author checklist

Before I request a review:

- [ ] ~I have included appropriate tests.~
- [ ] ~I have populated the deploy-preview with relevant test data.~
- [ ] ~I have tagged a UI/UX designer for review if this PR includes UI/UX changes.~


Kaustubh9031 (approved):
LGTM. Testing can be done after pre-release.

m4manjesh:
@ShushN Is there a reason you didn't create a release branch for this PR?

Kaustubh9031:
> @ShushN Is there a reason you didn't create a release branch for this PR?

Kaustubh9031:
@m4manjesh Matt suggested to keep the target branch as `main` as of now.

mattcorner (changes requested):
The fix made for this PR only resolves the issue when we have links, but we can truncate other text content. The three dots should match the colour of the content which is not always `text.primary`. You fix makes all text content `text.primary` even when it shouldn't be.

See all the examples on main vs this branch:

https://main--666c1309face6e7c27f20e04.chromatic.com/?path=/docs/general-truncatedtooltip--docs

https://666c1309face6e7c27f20e04-bhrkkjsetv.chromatic.com/?path=/docs/general-truncatedtooltip--docs

Kaustubh9031:
@mattcorner @ShushN I agree with Matt. When the child component of `NoWrapTypography` has color other than `primary`, it doesn't align with the `primary` color of the overflow `ellipsis`.
A potential solution would be to get the `color` of the `children` in the `NoWrapTypography` and then pass it to the `TruncatedTooltip`. This color then can be used inside `sx` prop in the `TruncatedTooltip`.

**e.g.**
export default function NoWrapTypography({
  children,
  sx,
  variant
}: NoWrapTypographyProps) {
  const color = React.isValidElement(children) && children.props?.sx?.color ? children.props.sx.color : "inherit";
  return (
    <TruncatedTooltip
      component={Typography}
      variant={variant}
      sx={sx}
      color={color}
    >
      {children}
    </TruncatedTooltip>
  );
}

src/TruncatedTooltip/TruncatedTooltip.tsx
```diff
@@ -26,9 +27,20 @@ const TruncatedTooltip = <T extends React.ElementType = "span">({
...
+      // Fetch the computed styles dynamically
+      const firstChild = textElementRef.current.children[0] as HTMLElement;
+      const color = window.getComputedStyle(firstChild).color;
+      setComputedColor(color);
+    }
+  }, []);
   // State to determine if the tooltip should show.
   const [open, setOpen] = useState(false);
 
+  console.log(computedColor);
```
Kaustubh9031, line 42: remove console statement

src/TruncatedTooltip/TruncatedTooltip.tsx
```diff
@@ -26,6 +27,15 @@ const TruncatedTooltip = <T extends React.ElementType = "span">({
 }: TruncatedTooltipProps<T>) => {
   // Ref to the text element.
   const textElementRef = useRef<HTMLInputElement | null>(null);
+  const [computedColor, setComputedColor] = useState<string>("");
+  useEffect(() => {
+    if (textElementRef.current?.children.length) {
+      // Fetch the computed styles dynamically
```
Kaustubh9031, line 33: The comment does not align with its adjacent line of code. Please write appropriate code comments.

Kaustubh9031 (changes requested):
Request changes

Kaustubh9031 (approved):
LGTM. I have tested the changes and it's working as expected.

@mattcorner Please do a final review.

mattcorner (changes requested):
image

It doesn't work as expected for "Link using children" in dark mode. 

Keep in mind that this component can be rendered with the `component` prop, or the `children`. The issues only seems to occur for the `children` case as per the bug report.

Kaustubh9031:
> image
> It doesn't work as expected for "Link using children" in dark mode.
> 
> Keep in mind that this component can be rendered with the `component` prop, or the `children`. The issues only seems to occur for the `children` case as per the bug report.

useEffect(() => {
    if (textElementRef.current?.children.length) {
      // Get the child element of the reference node
      const firstChild = textElementRef.current.children[0] as HTMLElement;
      // Retrieve the computed color style of the child element
      const color = window.getComputedStyle(firstChild).color;
      setComputedColor(color);
    }
  }, [computedColor]);

Changing the dependency array works but need to refresh the page while switching between light and dark mode. Please let know if we can proceed with this.

Kaustubh9031:
import { useTheme } from "@mui/material";


const theme = useTheme(); // Access the current theme
  // Ref to the text element.
  const textElementRef = useRef<HTMLInputElement | null>(null);
  const [computedColor, setComputedColor] = useState<string>("");
  useEffect(() => {
    if (textElementRef.current?.children.length) {
      // Get the child element of the reference node
      const firstChild = textElementRef.current.children[0] as HTMLElement;
      // Retrieve the computed color style of the child element
      const color = window.getComputedStyle(firstChild).color;
      setComputedColor(color);
    }
  }, [computedColor, theme]);

Using the useTheme hook from MUI and changing the dependency array to [computedColor, theme] works perfectly.

mattcorner:
Hey @ShushN @Kaustubh9031. I've had a play around and the underlying `TruncatedTooltip` component itself already does exactly what we need without any changes. It is just being used incorrectly.

I have updated the stories to better reflect how the component should be used. When we want to truncate a link, then we should do something like:

```tsx
<TruncatedTooltip 
    component={Link},
    href: "This/can/take/an/href/prop",
    variant: "body1"
>
   Some long text that will be truncated
<TruncatedTooltip>
```

The reason that this works is the truncation styling gets applied to the component that is passed in as `component`. Hence the ellipsis colour matches the colour of the element being rendered. No client-side JS needed to manipulate the styles.

We then have the problem that we have two incredibly similar components in `NoWrapTypography` and `TruncatedTooltip`.  `NoWrapTypography` was the original component and was never intentionally written to support Links. `TruncatedTooltip` was written to be flexible and was intended to support Links. Unfortunately `NoWrapTypography` is the more widely used component in VIRTO. 

At this point `NoWrapTypography` is a simple wrapper around `TruncatedTooltip` which reduces the functionality and is easily misused as shown in the bug ticket. I suggest we delete the `NoWrapTypography` component from RUI. This would be a breaking change and so I would target suggest targeting RUIv11 alongside the MUIv6 changes.

We would then need to go and replace every usage of `NoWrapTypography` in VIRTO with the correct usage of `TruncatedTooltip`. We would also need to fix any existing [incorrect usage](https://github.com/IPG-Automotive-UK/virto/blob/e994191f147cf9d80a906a5cf529a973da9e89ee/apps/scene/app/src/components/TableView/RenderRoadNameCell.tsx#L15) of the `NoWrapTypography` component.

cc @lukemojo I know you wrote `TruncatedTooltip`. Do you agree with the above proposal?

lukemojo:
> cc @lukemojo I know you wrote `TruncatedTooltip`. Do you agree with the above proposal?

@mattcorner Apologies for missing this yesterday. Yes I agree with the proposal. The NoWrapTypography component was maintained to prevent a breaking change across Virto but it doesn't make much sense for both components to exist if its causing problems. It will result in slightly more complexity in implementation due to the flexibility. 

Storybook should have all the relevant examples. Note that once you pass the component prop, TruncatedTooltip will then accept all relevant component props. I.e. component={Typography} will allow the component to accept all MUI Typography props. 

The target branch for this will currently be v11 which I will update now.

Kaustubh9031 (changes requested):
@ShushN Recently some changes have been made as part of MUI v6 migration which is reverted in this PR. Please take care of this and commit again.

src/Filter/CheckboxFilter/CheckboxFilter.tsx
```diff
@@ -84,9 +84,9 @@ function Option(
         icon={<CheckBoxOutlineBlank fontSize="small" />}
         value={option}
       />
-      <NoWrapTypography data-testid="filter-option-label">
+      <TruncatedTooltip data-testid="filter-option-label">
```
Kaustubh9031, line 87: pass `typography` as component prop

src/Filter/LabelFilter/LabelFilter.tsx
```diff
@@ -120,10 +121,10 @@ function Option(
         }}
       />
       <Stack direction="column" flexGrow={1} overflow="hidden">
-        <NoWrapTypography>{option.name}</NoWrapTypography>
-        <NoWrapTypography variant="caption">
+        <TruncatedTooltip>{option.name}</TruncatedTooltip>
```
Kaustubh9031, line 124: pass `typography` as component prop

src/LabelSelector/LabelSelector.jsx
```diff
@@ -219,10 +219,10 @@ export default function LabelSelector({
                   overflow: "hidden"
                 }}
               >
-                <NoWrapTypography>{option.name}</NoWrapTypography>
-                <NoWrapTypography variant="caption">
+                <TruncatedTooltip>{option.name}</TruncatedTooltip>
```
Kaustubh9031, line 222: pass typography as component prop

src/Filter/CheckboxFilter/CheckboxFilter.tsx
Kaustubh9031, line 87: pass `component={Typography}`

src/Filter/LabelFilter/LabelFilter.tsx
Kaustubh9031, line 124: pass `component={Typography}`

src/LabelSelector/LabelSelector.jsx
Kaustubh9031, line 222: pass `component={Typography}`

src/LabelSelector/LabelSelector.jsx
```diff
@@ -219,10 +219,10 @@ export default function LabelSelector({
...
+                <TruncatedTooltip variant="caption">
```
Kaustubh9031, line 223: pass `component={Typography}`

src/TruncatedTooltip/TruncatedTooltip.tsx
```diff
@@ -26,6 +26,7 @@ const TruncatedTooltip = <T extends React.ElementType = "span">({
 }: TruncatedTooltipProps<T>) => {
   // Ref to the text element.
   const textElementRef = useRef<HTMLInputElement | null>(null);
+
```
Kaustubh9031, line 29: Remove this extra line

Kaustubh9031 (changes requested):
Please make the changes based on feedback.

Kaustubh9031 (approved):
LGTM

PR #1024, opened 2025-01-15
VE-265 ESM & CJS Builds

dmbarry86:
Contributes to [VE-265](https://sce.myjetbrains.com/youtrack/issue/VE-265/React-Router-v7-concept)

## Changes

Previously we were building and publishing ESM only. I've now updated so we build and publish both ESM and CJS. This can support SSR environments that need CJS, e.g. React Route V7 with Vite. This is a similar concept to how MUI build and publish where both CJS and ESM are available.

* Created build scripts for CJS and ESM. The builds are output in dist/cjs and dist/esm. There are 2 new tsconfig files that build on top of the base tsconfig; one for each build env. The "production" tsconfig has now been dropped and replaced by the base tsconfig and the esm tsconfig.
* Updated main build script to call both of the build scripts mentioned above.
* Updated the copy-files script to copy static assets to both CSM and ESM build directories.
* Added main, module and export declarations for ESM and CJS to point to the approriate build directories in dist.

## Dependencies

N/A

## UI/UX

N/A

## Testing notes

Run a build and check that ESM and CJS code is generated accordingly in dist/esm and dist/cjs. The ESM build is effectively what we had before.

## Author checklist

Before I request a review:

- [ ] ~I have tested the changes in Docker / a deploy-preview.~
- [ ] ~I have included appropriate tests.~
- [ ] ~I have populated the deploy-preview with relevant test data.~
- [ ] ~I have tagged a UI/UX designer for review if this PR includes UI/UX changes.~


tsconfig.json
```diff
@@ -2,18 +2,29 @@
...
+  },
+  "include": ["src/**/*.tsx", "src/**/*.ts", "src/**/*.jsx", "src/**/*.js"],
+  "exclude": [
+    "src/**/*.test.**",
+    "src/**/*.spec.**",
+    "src/**/*.stories.**",
+    "dist",
+    "vitest-setup.ts",
+    "vite.config.ts",
+    "vite.config.ts"
```
mattcorner, line 28: The reason that we had a dev + production tsconfig before was that I can now introduce type errors in tests and stories. While VSCode does highlight them to me, `pnpm run type-check` does not because the tsconfig tells typescript to ignore these files.

mattcorner (changes requested):
LGTM, just an issue with type-checking non-production files

tsconfig.json
dmbarry86: I'm not sure this is true. I can introduce a type error on main in one of the stories and it's not picked up by type-check.

image

If I actually implement what you are talking about, we have 141 existing issues.

image

dmbarry86: I propose a separate ticket to cover this existing issue so we don't hold up the start of the DATA migration. If it was just a couple of errors we could absorb it now but 141 is not a 5 minute job. I'm happy to look into it once the DATA work is underway.

dmbarry86: https://sce.myjetbrains.com/youtrack/issue/VE-290/Type-checks-for-tests-and-stories-in-RUI

mattcorner (approved):

PR #1076, opened 2025-03-18
TD-3479 Vehicle Selector Errors

lukemojo:
Closes [TD-3479](https://sce.myjetbrains.com/youtrack/issue/TD-3479/VehicleSelector-error-handling)

## Changes

Ensures errors appear onBlur. Adds a prop to force the validation of all fields

- Add a "validate" prop to validate active fields
- Ensure fields are validated on blur
- When validate prop is true, set all error states

## Testing notes

In storybook, check that empty fields show an error on blur by default. Check that an error appears with the validate prop is set to true (there is a validate story). Check that errors disappear when the field has a valid value.

## Author checklist

Before I request a review:

- ~[x] I have populated the deploy-preview with relevant test data.~
- [x] I have tagged a UI/UX designer for review if this PR includes UI/UX changes.

### DIFF
--- a/src/VehicleSelector/VehicleSelector.stories.tsx
+++ b/src/VehicleSelector/VehicleSelector.stories.tsx
@@ -327,3 +327,53 @@ export const WithExternalErrors = {
 
   render: Template
 };
+
+export const InitWithError = {
+  args: {
+    flexDirection: "column",
+    flexWrap: "nowrap",
+    gates: [],
+    validateOnInit: true,
+    value: [],
+    variants: [
+      {
+        _id: "1",
+        modelYear: "2015",
+        projectCode: "911",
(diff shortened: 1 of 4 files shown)


m4manjesh (changes requested):
Do we still need the `errors` prop? Nothig happens when i set the `validate` prop to false when there is no value. Do we expect any field validation without `validate` prop true?

lukemojo:
> Do we still need the `errors` prop? Nothig happens when i set the `validate` prop to false when there is no value. Do we expect any field validation without `validate` prop true?

No, I've removed the type, that was an oversight. If you set validate to false nothing should happen aside from error suppression, when you set validate to true it should validate all the active fields. Regardless of the validate prop boolean, all fields should still validate onBlur.

PR #1079, opened 2025-03-30
TD-3527 Update VersionChip to support single digits

karthikraj29:
Closes  [TD-3527](https://sce.myjetbrains.com/youtrack/issue/TD-3527/Update-VersionChip-to-Support-Whole-Numbers)

## Changes

Currently, the VersionChip only supports decimal formats like 1.2 ("major.minor"). Update the function to also allow whole numbers like 1.

1. Updated the regex to allow both whole numbers (1) and decimal numbers (1.2).

2. Modified the function to handle cases where the minor version is missing (defaulting it to "0" if needed).

## Dependencies

NA

## UI/UX

NA

## Testing notes

Run `pnpm Storybook` , Go to `VersionChip`, then test the below .

1. **Whole Number Support**  
   - Test with `1` → Should be accepted.

2. **Decimal Number Support**  
   - Test with `1.0` → Should be accepted.
   - Test with `1.2` → Should be accepted as is.  

4. **Invalid Format Handling**  
   - Test with `1.2.3` →  Should not be accepted and log a warning .
   - Test with `abc` →  Should not be accepted and log a warning.  

## Author checklist

Before I request a review:

- [x] I have tested the changes in ~Docker / a deploy-preview~ stroryBook.
~[ ] I have populated the deploy-preview with relevant test data.~
~[ ] I have tagged a UI/UX designer for review if this PR includes UI/UX changes.~

### DIFF
--- a/src/VersionChip/VersionChip.stories.tsx
+++ b/src/VersionChip/VersionChip.stories.tsx
@@ -26,7 +26,7 @@ export const Default = {
   render: Template
 };
 
-export const WholeNumber = {
+export const SingleDigit = {
   args: {
     selected: false,
     version: "1"
@@ -44,7 +44,7 @@ export const MinorVersion = {
   render: Template
 };
 
-export const WholeNumberSelected = {
+export const SingleDigitSelected = {
   args: {
     selected: true,
     version: "1"
--- a/src/VersionChip/VersionChip.stories.tsx
+++ b/src/VersionChip/VersionChip.stories.tsx
@@ -26,10 +26,28 @@ export const Default = {
   render: Template
 };
 
+export const WholeNumber = {
+  args: {
+    selected: false,
+    version: "1"
+  },
+
+  render: Template
+};
+
 export const MinorVersion = {
   args: {
     selected: false,
     version: "1.1"
+  },
+
+  render: Template
+};
+
+export const WholeNumberSelected = {
+  args: {
+    selected: true,
+    version: "1"
   },
 
   render: Template
--- a/src/VersionChip/VersionChip.test.tsx
+++ b/src/VersionChip/VersionChip.test.tsx
@@ -1,6 +1,6 @@
 import "@testing-library/jest-dom";
 
-import { afterAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { render, screen, waitFor } from "@testing-library/react";
 
 import React from "react";
@@ -58,24 +58,13 @@ describe("VersionChip", () => {
     expect(styles.border).toBe("1px solid rgba(0, 0, 0, 0.23)");
   });
 
-  // mock console.warn to suppress warnings in test output
-  const consoleWarnMock = vi
-    .spyOn(console, "warn")
-    .mockImplementation(() => {});
-
-  // Clear mock before each test
-  beforeEach(() => {
-    process.env.NODE_ENV = "development";
-    consoleWarnMock.mockClear();
-  });
-
-  // restore console.warn after all tests
-  afterAll(() => {
-    consoleWarnMock.mockRestore();
-  });
-
-  // Check if warning is logged for invalid version format
+  // check if warning is logged for invalid version format
   it("logs a warning for an invalid version format", async () => {
+    // mock console.warn to detect if a warning is logged
+    const consoleWarnMock = vi
+      .spyOn(console, "warn")
+      .mockImplementation(() => {});
+
     render(<VersionChip version="invalid-version" />);
 
     await waitFor(() => {
@@ -83,5 +72,6 @@ describe("VersionChip", () => {
         'Invalid version format: "invalid-version". Expected format is "<major>" or <major>.<minor>" (e.g., "1" or "1.0").'
       );
     });
+    consoleWarnMock.mockRestore();
   });
 });
--- a/src/VersionChip/VersionChip.test.tsx
+++ b/src/VersionChip/VersionChip.test.tsx
@@ -6,9 +6,19 @@ import VersionChip from "./VersionChip";
 
 describe("VersionChip", () => {
   // test that the version chip renders
+  it("renders when version is single digit", () => {
+    render(<VersionChip version="1"></VersionChip>);
+    expect(screen.getByText("1")).toBeInTheDocument();
+  });
+  // test that the version chip renders
   it("renders VersionChip", () => {
     render(<VersionChip version="1.0"></VersionChip>);
     expect(screen.getByText("1.0")).toBeInTheDocument();
+  });
+  // check single digit major version has correct svg
+  it("renders major svg", () => {
+    render(<VersionChip version="1"></VersionChip>);
+    expect(screen.getByTestId("LayersIcon")).toBeInTheDocument();
   });
   // check major version has correct svg
   it("renders major svg", () => {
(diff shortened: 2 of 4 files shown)


src/VersionChip/VersionChip.stories.tsx
```diff
@@ -26,6 +26,15 @@ export const Default = {
   render: Template
 };
 
+export const WholeNumber = {
```
AronPetrauskas, line 29: I would am more of a fan of the vocabulary of `single digit` vs `whole number` in our current context (we allow 1 and 1.1) the concept of a whole number makes sense, but if the regex is extended in the future to (1.1.1) this is not a mathematically correct number anymore and the concept of a whole number is not really relevant anymore here.

```suggestion
export const SingleDigit= {
```

src/VersionChip/VersionChip.stories.tsx
```diff
@@ -35,6 +44,15 @@ export const MinorVersion = {
   render: Template
 };
 
+export const WholeNumberSelected = {
```
AronPetrauskas, line 47:
```suggestion
export const SingleDigitSelected = {
```

src/VersionChip/VersionChip.test.tsx
```diff
@@ -5,11 +5,21 @@ import React from "react";
 import VersionChip from "./VersionChip";
 
 describe("VersionChip", () => {
+  // test that the version chip renders
+  it("renders when version is single digit", () => {
```
AronPetrauskas, line 9:
```suggestion
  it("renders when version is a single digit", () => {
```

src/VersionChip/VersionChip.test.tsx
```diff
@@ -5,11 +5,21 @@ import React from "react";
...
+    render(<VersionChip version="1"></VersionChip>);
+    expect(screen.getByText("1")).toBeInTheDocument();
+  });
   // test that the version chip renders
   it("renders VersionChip", () => {
     render(<VersionChip version="1.0"></VersionChip>);
     expect(screen.getByText("1.0")).toBeInTheDocument();
   });
+  // check single digit major version has correct svg
+  it("renders major svg", () => {
```
AronPetrauskas, line 19:
```suggestion
  it("renders major svg when version is a single digit", () => {
```

src/VersionChip/VersionChip.tsx
```diff
@@ -61,23 +61,25 @@ const VersionChip = ({ version, selected = false }: VersionChipProps) => {
...
 /**
- * A function that returns the minor version, will return `undefined` if format doesn't match number.number
+ * A function that returns the minor version,
+ * Returns `undefined` if the format is invalid.
+ * Accepts formats like "1" or "1.0" but not "1.1.1".
  */
 const getMinorVersion = (version: string) => {
-  // check format is number.number
-  const regex = /^\d+\.\d+$/;
+  // allow whole numbers or "major.minor" format (e.g., "1" or "1.0")
```
AronPetrauskas, line 69:
```suggestion
  // allow single digits or "major.minor" format (e.g., "1" or "1.0")
```

src/VersionChip/VersionChip.test.tsx
AronPetrauskas, line 1: Looking at the code coverage there are some uncovered cases. This has not been introduced by you, but rather by me in the past (sorry). I think we should add three more tests to achieve 100% code coverage.

1. Version chip NOT selected, check that border and chip color are that which we expect when the chip is NOT selected.
2. Version chip IS selected, check that border and chip color are that which we expect when the chip IS selected.
3. Input is invalid, check that chip does NOT render and that a console.warn is logged.

AronPetrauskas (changes requested):
Component is perfectly functional, just some comments for consistent vocabulary and suggestion for extra tests to increase code coverage.

src/VersionChip/VersionChip.test.tsx
```diff
@@ -35,4 +37,51 @@ describe("VersionChip", () => {
...
   });
+
+  // check background color when the `selected` prop is true
+  it("applies correct background color when `selected` is true", () => {
+    render(<VersionChip version="1.0" selected={true} />);
+    const chipElement = screen.getByTestId("version-chip");
+    const styles = window.getComputedStyle(chipElement);
+    // passing resolved color from alpha(theme.palette.info.main, 0.12) and theme.palette.primary.main
+    expect(styles.backgroundColor).toBe("rgba(2, 136, 209, 0.12)");
+    expect(styles.border).toBe("1px solid #1976d2");
```
AronPetrauskas, line 48: Please be consistent with the color encoding, you've used rgba mostly, lets stick to that here too
```suggestion
    expect(styles.border).toBe("1px solid rgba(25, 118, 210, 1)");
```

src/VersionChip/VersionChip.test.tsx
```diff
@@ -35,4 +37,51 @@ describe("VersionChip", () => {
...
+  it("applies correct background color when `selected` is false", () => {
+    render(<VersionChip version="1.0" selected={false} />);
+    const chipElement = screen.getByTestId("version-chip");
+    const styles = window.getComputedStyle(chipElement);
+    // resolved color from alpha(theme.palette.divider, 0.23) and theme.palette.background.default
+    expect(styles.backgroundColor).toBe("rgb(255, 255, 255)");
+    expect(styles.border).toBe("1px solid rgba(0, 0, 0, 0.23)");
+  });
+
+  // mock console.warn to suppress warnings in test output
```
AronPetrauskas, line 61:
```suggestion
  // mock console.warn to detect if a warning is logged
```

src/VersionChip/VersionChip.test.tsx
```diff
@@ -35,4 +37,51 @@ describe("VersionChip", () => {
...
+  const consoleWarnMock = vi
+    .spyOn(console, "warn")
+    .mockImplementation(() => {});
+
+  // Clear mock before each test
+  beforeEach(() => {
+    process.env.NODE_ENV = "development";
```
AronPetrauskas, line 68: Why is this line here? When we are testing the node environment should be "test". You don't need to worry about it as it is set upstream, but remove this line.
```suggestion
```

src/VersionChip/VersionChip.test.tsx
```diff
@@ -35,4 +37,51 @@ describe("VersionChip", () => {
...
+    consoleWarnMock.mockClear();
+  });
+
+  // restore console.warn after all tests
+  afterAll(() => {
+    consoleWarnMock.mockRestore();
+  });
+
+  // Check if warning is logged for invalid version format
+  it("logs a warning for an invalid version format", async () => {
```
AronPetrauskas, line 78: There is only this one test that needs the mocks you have setup, please move the mocks inside the test. Contact me if you need help or more context please.

AronPetrauskas (changes requested):
Some suggestions on how to clean-up the newly added test.

Code coverage is perfect now, good job. 
image

src/VersionChip/VersionChip.test.tsx
karthikraj29: I tried that before, however we'll get this error, as we receiving the style in hex format
image

AronPetrauskas: In this case proceed with the hex value

src/VersionChip/VersionChip.test.tsx
karthikraj29: The warning only logs in `development` mode, but tests run in `test` mode, so it doesn't trigger. Setting `NODE_ENV` to `development` ensures the test works. Any other alternative?

AronPetrauskas (dismissed):
Nice work. LGTM!

